PR #855, opened 2018-06-26
Banned user reviews

lyyder:
Currently showing reviews from banned users breaks as the data provided for those users by the API is limited. This PR solves this problem by showing a placeholder name in place of the author display name in such cases. Also the background of the banned user avatar is changed to match the marketplace color theme.

screen shot 2018-06-26 at 12 50 18

### DIFF
--- a/src/components/IconBannedUser/IconBannedUser.css
+++ b/src/components/IconBannedUser/IconBannedUser.css
@@ -1,11 +1,11 @@
 @import '../../marketplace.css';
 
 .backgroundLight {
-  stop-color: var(--marketplaceColorLight);
+  stop-color: var(--bannedColorLight);
 }
 
 .backgroundDark {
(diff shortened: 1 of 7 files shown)


src/components/IconBannedUser/IconBannedUser.css
```diff
@@ -1,7 +1,11 @@
 @import '../../marketplace.css';
 
-.backgroundFill {
-  fill: var(--failColor);
+.backgroundLight {
+  stop-color: var(--marketplaceColorLight);
+}
+
+.backgroundDark {
+  stop-color: var(--marketplaceColor);
```
Gnito: I'd add more variables to marketplace.css

```css
--bannedColorLight: var(--marketplaceColorLight);
```

```
stop-color: var(--bannedColorLight);
```

PR #861, opened 2018-07-05
Fix clearing tx initial message before the page change is made

lyyder:
Fixes a UI bug on checkout page where the initial message input is cleared slightly before the page changes to the tx page.

__Before:__
message_blink_before

__After:__
message_blink_after


lyyder:
Now the state is not cleared during submit. A question is that if some values, like `token`, should be cleared on submit.

PR #862, opened 2018-07-06
User card layout

lyyder:
In a certain viewport with a long user name or translations, _Edit profile_ link used to overlap with the heading in `UserCard` (actually the link was rendered outside the `UserCard` component):

desktop_before

No the edit profile link has been moved inside the `UserCard` component and it does not overlap the heading:

desktop_after

Also previously in mobile view the edit profile link used to raise above the whole user card:

mobile_before

Now it's placed in place of the _Contact_ link if the user of the card is logged in:

mobile_after

### DIFF
--- a/src/components/UserCard/UserCard.css
+++ b/src/components/UserCard/UserCard.css
@@ -114,3 +114,8 @@
     display: none;
   }
 }
+
+.contact {
+  @apply --marketplaceH4FontStyles;
+  margin: 0;
+}
--- a/src/components/UserCard/UserCard.css
+++ b/src/components/UserCard/UserCard.css
@@ -99,8 +99,7 @@
   @apply --marketplaceH4FontStyles;
   display: none;
   margin-bottom: 0;
-
-  /* Align to same baseline as the "Your host" heading */
+  white-space: nowrap;
 
   @media (--viewportMedium) {
     display: inline;
@@ -110,8 +109,6 @@
 .editProfileMobile {
   @apply --marketplaceH4FontStyles;
   display: inline;
-
-  /* Align to same baseline as the "Your host" heading */
 
   @media (--viewportMedium) {
     display: none;
--- a/src/components/UserCard/UserCard.css
+++ b/src/components/UserCard/UserCard.css
@@ -53,6 +53,7 @@
   display: none;
 
   @media (--viewportMedium) {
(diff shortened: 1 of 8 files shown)


src/components/UserCard/UserCard.js
```diff
@@ -83,20 +83,37 @@ const UserCard = props => {
     [css.withBioMissingAbove]: !hasBio,
   });
 
-  const hideContact = currentUser && isCurrentUser;
-  const separator = hideContact ? null : <span className={css.linkSeparator}>•</span>;
-  const contact = hideContact ? null : (
+  const separator = isCurrentUser ? null : <span className={css.linkSeparator}>•</span>;
```
Gnito: The intended behaviour is:
- unauthenticated user: `• contact`
- authenticated user, but not author: `• contact`
- authenticated user, author: `• edit`

If I understood this right, the separator should be printed always (no need for ifs anymore)

Also, `{contact}` and `{editProfileMobile}` are alternatives for each others. So, for the clarity: 
 => `{isCurrentUser ? editProfileMobile : contact}`

src/components/UserCard/UserCard.css
```diff
@@ -64,16 +65,52 @@
...
+.headingRow {
+  display: flex;
+  flex-direction: column;
+
+  @media (--viewportMedium) {
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: baseline;
+  }
+}
```
Gnito: Could you check that the row is always divisible by 6px (and 8px on the desktop layout)? If not, add some padding to top and bottom so that it's size will match that requirement and texts are baselined nicely.

We should be more pedantic with our baseline rules so that there's less need for random margins between elements (i.e. `margin-top: 12px` makes sense, but `13px` is wrong -> some other component has wrong height).

lyyder: @Gnito The component is now internally in baseline. However, I'll see if the css definitions could be improved a bit more to have more logical values regarding the baseline definitions.

src/components/UserCard/UserCard.js
lyyder: Almost, but as seen in the second screenshot from the top, in the desktop view the edit profile link is on the right and the separator is hidden.

PR #863, opened 2018-07-09
Add clear errors action

lyyder:
Add an action that can be used to clear errors on load data instead of
clearing the whole state.

Fixes passing an initial message sending error to the tx page. The initial message sending error is cleared when a new send message request is fired. Downside to this is that the initial message error is not cleared between component loads in case sending the message is not tried again.

### DIFF
--- a/src/containers/TransactionPage/TransactionPage.duck.js
+++ b/src/containers/TransactionPage/TransactionPage.duck.js
@@ -27,6 +27,8 @@ const CUSTOMER = 'customer';
 
 export const SET_INITAL_VALUES = 'app/TransactionPage/SET_INITIAL_VALUES';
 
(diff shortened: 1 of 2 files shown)


src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -470,7 +483,7 @@ export const loadData = params => dispatch => {
   const txId = new UUID(params.id);
 
   // Clear the send error since the message form is emptied as well.
-  dispatch(setInitialValues({ sendMessageError: null, sendReviewError: null }));
+  dispatch(clearErrors());
```
Gnito: 👍 
Did you check if any other files had setIntialValues called inside loadData?

lyyder: @Gnito good question. This seems to be the only loadData implementation that does this.

PR #866, opened 2018-07-10
Improve tx page state management

lyyder:
The state management in the `loadData` function in `TransactionPage.duck` has been somewhat problematic and this PR aims to improve it.

Now `loadData` checks if a `transactionRef` is found from state. It is set when a tx fetch succeeds. If a `transactionRef` is found it means the state holds data from a previous page load and the state is cleared. If the ref can not be found it means that the page has not been loaded or that the `setInitialValues` has been invoked for example with an initial message sending error information from `CheckoutPage`.

### DIFF
--- a/src/containers/TransactionPage/TransactionPage.duck.js
+++ b/src/containers/TransactionPage/TransactionPage.duck.js
@@ -1,4 +1,5 @@
 import pick from 'lodash/pick';
+import pickBy from 'lodash/pickBy';
 import { types as sdkTypes } from '../../util/sdkLoader';
 import { isTransactionsTransitionInvalidTransition, storableError } from '../../util/errors';
 import {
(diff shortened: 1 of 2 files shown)


src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -479,11 +472,16 @@ export const sendReview = (role, tx, reviewRating, reviewContent) => (dispatch,
...
-
-  // Clear the send error since the message form is emptied as well.
-  dispatch(resetState());
+  const state = getState().TransactionPage;
+  const txRef = state.transactionRef;
+
+  // In case a transaction reference is found from a previous
+  // data load -> clear the state. Otherwise keep the non-null
+  // values which may have been set from a previous page.
+  const initialValues = txRef ? {} : pickBy(state, v => !!v);
```
Gnito: Arrays are a problem here.
`!![]` => true

`v => (Array.isArray(v) && !!v.length) || !!v`

lyyder: @Gnito I thought about that and reasoned that if the state has en empty array it can be passed as an initial value as our initial value for `messages`, which seems to be the only array, is `[]`.

However, what you proposed would be more explicit. Also empty objects are truthy too and for example there's the `transactionRef` which is an object and should be overridden with the value from `initialState` in case it is somehow has been set to `{}`.

lyyder: Also, passing `[]` to

`(Array.isArray(v) && !!v.length) || !!v`

resolves to `true`

PR #867, opened 2018-07-12
Add sessiontoken and sparse fields

Gnito:
This is a reaction to Google Maps pricing change: 
- Added sessionToken to Autocomplete calls (otherwise calls are prices separately)
- Use sparse fields for place details call (otherwise all the data is returned and it costs more)

### DIFF
--- a/src/components/LocationAutocompleteInput/LocationAutocompleteInput.js
+++ b/src/components/LocationAutocompleteInput/LocationAutocompleteInput.js
@@ -152,6 +152,10 @@ class LocationAutocompleteInput extends Component {
     // Ref to the input element.
     this.input = null;
 
+    // Current sessionToken used to combine autocomplete calls with place details call
+    // This reduces Google Maps pricing.
+    this.autocompleteSessionToken = null;
+
     this.changeHighlight = this.changeHighlight.bind(this);
     this.selectItem = this.selectItem.bind(this);
     this.selectItemIfNoneSelected = this.selectItemIfNoneSelected.bind(this);
@@ -213,6 +217,7 @@ class LocationAutocompleteInput extends Component {
     this.setState({ highlightedIndex: -1 });
 
     if (!newValue) {
+      this.autocompleteSessionToken = null;
       // No need to fetch predictions on empty input
       return;
     }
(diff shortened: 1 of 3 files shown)


src/components/LocationAutocompleteInput/LocationAutocompleteInput.js
```diff
@@ -213,6 +217,7 @@ class LocationAutocompleteInput extends Component {
     this.setState({ highlightedIndex: -1 });
 
     if (!newValue) {
+      this.autocompleteSessionToken = null;
```
bladealslayer: Does the token need to be cleared out when user clears input when the place details have not been retrieved? E.g. input "Hel" -> delete all -> input "Espoo" -> click the item? Is this one or two sessions?

src/util/googleMaps.js
```diff
@@ -47,9 +47,10 @@ export const getPlaceDetails = (placeId, sessionToken) =>
     const serviceStatus = window.google.maps.places.PlacesServiceStatus;
     const el = document.createElement('div');
     const service = new window.google.maps.places.PlacesService(el);
+    const fields = ['address_component', 'formatted_address', 'geometry', 'place_id'];
```
lyyder: Isn't the `place_id` known already?

Gnito: Yes, but it doesn't hurt to have that returned also (it's part of the basic fields => same pricing). Anyway, id's included to the return object makes them easier to be reused in Redux store if that kind of feature is needed.

src/components/LocationAutocompleteInput/LocationAutocompleteInput.js
Gnito: @bladealslayer that's what I understood from docs, but I'm not sure. Should be asked from Google guys. Those docs just say that if the same session token is used in several sessions/occasions, all the calls are handled individually -> it cancels the price reduction of the token.

I assume that there's an undisclosed lifetime on those tokens - but since it's undisclosed, I assume that the desired behavior is that one character search "sequence" group is one session: "h", "he", "hel" -> "Helsinki".

Gnito: https://developers.google.com/maps/documentation/javascript/places-autocomplete#session_tokens
"Warning: Be sure to pass a unique session token for each new session. Using the same token for more than one session will result in each request being billed individually. Note that the Autocomplete Widget handles sessions automatically (you don't need to do anything additional)."

PR #868, opened 2018-07-17
Fetch availability data

lyyder:
Adds availability data fetching for listing page on page load.


src/containers/ListingPage/ListingPage.duck.js
```diff
@@ -190,5 +229,19 @@ export const loadData = (params, search) => dispatch => {
...
+  // in the future due to Stripe limitations
+  const start = moment().toDate();
+  const end = moment()
+    .add(config.dayCountAvailableForBooking, 'days')
+    .toDate();
+  const timeSlotsParams = { listingId, start, end };
+
+  return Promise.all([
+    dispatch(showListing(listingId)),
+    dispatch(fetchTimeSlots(timeSlotsParams)),
```
Gnito: As discussed IRL, I think this call should be included when flagged somehow in config.js

lyyder:
Put time slots fetching behind a flag. I'll leave the error message handling along with a changelog entry to a future PR when the availability can be passed to the calendar components and this whole thing can be tied together.

PR #869, opened 2018-07-18
Listingpage static map

Gnito:
NOTE 1: fuzzy location needs some extra work otherwise every page refresh is going to change the static map URL.

NOTE 2: map marker is moved to public/static/icons folder to be available for Google's static map API (works in production - localhost shows an error on top right corner of the generated map since the googleapis.com can't fetch the image from localhost.)


src/components/Map/StaticMap.js
```diff
@@ -0,0 +1,177 @@
...
+
+    points.push([pLat.toFixed(6), pLng.toFixed(6)]);
+  }
+
+  return polyline.encode(points);
+};
+
+// Extract color from string. Given value should be either with '#' (e.g. #FFFFFF') or without it.
+const formatColorFromString = color => {
+  if (typeof color !== 'string' && /^#[0-9A-F]{6}$/i.test(color)) {
```
lyyder: What should the type be here if not string?

lyyder: Should it be `typeof color === 'string'`? Same thing also below.

src/util/contextHelpers.js
```diff
@@ -68,3 +68,110 @@ export const withViewport = Component => {
...
+ * @param {React.Component} Component to be wrapped by this HOC
+ * @param {Object} options pass in options like maxWidth and maxHeight.
+ *
+ * @return {Object} HOC component which knows its dimensions
+ */
+export const lazyLoadWithDimensions = (Component, options) => {
+  // The resize event is flooded when the browser is resized. We'll
+  // use a small timeout to throttle changing the viewport since it
+  // will trigger rerendering.
+  const THROTTLE_WAIT_MS = 200;
```
lyyder: 👍

PR #871, opened 2018-07-19
Improve listingpage lazy loading

Gnito:


src/util/contextHelpers.js
```diff
@@ -106,8 +110,8 @@ export const lazyLoadWithDimensions = (Component, options) => {
       window.addEventListener('orientationchange', this.handleWindowResize);
 
       this.defaultRenderTimeout = window.setTimeout(() => {
-        if (this.isElementInViewport()) {
-          this.handleWindowResize();
+        if (this.isElementNearViewport(0)) {
```
lyyder: `isElementNearViewport` doesn't seem to take a param.

src/util/contextHelpers.js
```diff
@@ -120,7 +124,9 @@ export const lazyLoadWithDimensions = (Component, options) => {
     }
 
     handleWindowResize() {
-      this.setDimensions();
+      if (this.isElementNearViewport(NEAR_VIEWPORT_MARGIN)) {
```
lyyder: `isElementNearViewport` doesn't seem to take a param, but it uses `NEAR_VIEWPORT_MARGIN` internally.

PR #873, opened 2018-07-23
Show available time slots in date input

lyyder:
Blocked dates use the same styling as the out of range ones.

screen shot 2018-07-23 at 19 57 47

### DIFF
--- a/src/components/FieldDateInput/DateInput.js
+++ b/src/components/FieldDateInput/DateInput.js
@@ -17,6 +17,7 @@ import classNames from 'classnames';
 import moment from 'moment';
 import config from '../../config';
 import { propTypes } from '../../util/types';
+import { dateFromAPIToLocalNoon } from '../../util/dates';
 
 import NextMonthIcon from './NextMonthIcon';
 import PreviousMonthIcon from './PreviousMonthIcon';
@@ -103,6 +104,16 @@ const defaultProps = {
   },
 };
 
+// Checks if time slot (propTypes.timeSlot) start time equals a day (moment)
+const timeSlotEqualsDay = (timeSlot, day) => {
+  // Time slots describe available dates by providing a start and
+  // an end date which is the following day. In the single date picker
+  // the start date is used to represent available dates.
+  const localStartDate = dateFromAPIToLocalNoon(timeSlot.attributes.start);
+
+  return isSameDay(day, moment(localStartDate));
+};
+
 class DateInputComponent extends Component {
   constructor(props) {
     super(props);
@@ -159,12 +170,7 @@ class DateInputComponent extends Component {
     const date = value && value.date instanceof Date ? moment(value.date) : initialMoment;
 
     const isDayBlocked = timeSlots
-      ? day => {
-          // Time slots describe available dates by providing a start and
-          // an end date which is the following day. In the single date picker
-          // the start date is used to represent available dates.
-          return !timeSlots.find(timeSlot => isSameDay(day, moment(timeSlot.attributes.start)));
-        }
+      ? day => !timeSlots.find(timeSlot => timeSlotEqualsDay(timeSlot, day))
       : () => false;
 
     const placeholder = placeholderText || intl.formatMessage({ id: 'FieldDateInput.placeholder' });
--- a/src/components/FieldDateInput/DateInput.js
+++ b/src/components/FieldDateInput/DateInput.js
@@ -160,6 +160,9 @@ class DateInputComponent extends Component {
 
     const isDayBlocked = timeSlots
       ? day => {
+          // Time slots describe available dates by providing a start and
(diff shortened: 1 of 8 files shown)


src/containers/ListingPage/ListingPage.js
```diff
@@ -497,7 +497,7 @@ ListingPageComponent.defaultProps = {
   showListingError: null,
   reviews: [],
   fetchReviewsError: null,
-  timeSlots: [],
+  timeSlots: null,
```
lyyder: The default `timeSlots` value on listing page level is changed to `null` so that it can be used to resolve if time slots are used. An empty array that would be passed to subcomponents would just render a blocked calendar.

src/components/FieldDateInput/DateInput.js
```diff
@@ -151,6 +157,12 @@ class DateInputComponent extends Component {
 
     const date = value && value.date instanceof Date ? moment(value.date) : initialMoment;
 
+    const isDayBlocked = timeSlots
+      ? day => {
+          return !timeSlots.find(timeSlot => isSameDay(day, moment(timeSlot.attributes.start)));
```
Gnito: Is the `timeSlot.attributes.start` always a start of a single day or could it be a range of days?
Either way, it could be good to have a short comment about response restrictions and rules? 
Another related response format question: is the returned day in UTC 00:00 timestamp or has the timezone difference added to it already?

Gnito: This might help: https://github.com/sharetribe/flex-template-web/blob/master/src/util/dates.js#L27

lyyder:
@Gnito addressed the comments, this is up for a new review.

src/components/FieldDateInput/DateInput.js
```diff
@@ -103,6 +104,16 @@ const defaultProps = {
...
 };
 
+// Checks if time slot (propTypes.timeSlot) start time equals a day (moment)
+const timeSlotEqualsDay = (timeSlot, day) => {
+  // Time slots describe available dates by providing a start and
+  // an end date which is the following day. In the single date picker
+  // the start date is used to represent available dates.
+  const localStartDate = dateFromAPIToLocalNoon(timeSlot.attributes.start);
+
+  return isSameDay(day, moment(localStartDate));
```
Gnito: Add also a check for `type: "time-slot/day"`

src/util/test-data.js
```diff
@@ -212,6 +213,33 @@ export const createReview = (id, attributes = {}, includes = {}) => {
...
+ */
+export const createTimeSlots = (startDate, numberOfDays) => {
+  const startTime = moment(startDate).startOf('day');
+
+  return Array.from({ length: numberOfDays }, (v, i) => i).map(i => {
+    return {
+      id: new UUID(i),
+      type: 'timeSlot',
+      attributes: {
+        start: moment(startTime)
```
Gnito: This needs to use UTC date - otherwise, customizers get errors if they run this (ie. their laptop) in different timezones than UTC.

src/components/FieldDateInput/DateInput.js
Gnito: Something like `ensureTimeSlot(timeSlot). attributes.type === 'time-slot/day'`

src/components/FieldDateInput/DateInput.js
```diff
@@ -106,12 +107,16 @@ const defaultProps = {
...
+  if (ensureTimeSlot(timeSlot).attributes.type === TIME_SLOT_DAY) {
+    // Time slots describe available dates by providing a start and
+    // an end date which is the following day. In the single date picker
+    // the start date is used to represent available dates.
+    const localStartDate = dateFromAPIToLocalNoon(timeSlot.attributes.start);
+
+    return isSameDay(day, moment(localStartDate));
+  } else {
+    return false;
+  }
```
Gnito: This works but it could be written shorter as 
```
const isDay = ensureTimeSlot(timeSlot).attributes.type === TIME_SLOT_DAY;
return isDay && isSameDay(day, moment(localStartDate))
```

src/util/test-data.js
Gnito: Hmm. This is only called from styleguide - so, it's not needed for test verification. I'm not sure if this will cause problems or not. (I was assuming that startDate should be something like `dateFromLocalToAPI(startDate)`, but moment's `startOf` function doesn't know about UTC start of day moment anyway...)

src/components/FieldDateInput/DateInput.js
lyyder: Looks elegant. One drawback though is that the `localStartDate` needs to be resolved with `dateFromAPItoLocalNoon` even though the time slot type wasn't of type day. But the `dateFromAPIToLocalNoon` should handle anything you throw at it.

PR #877, opened 2018-08-01
Reusable map container

Gnito:
This makes SearchMap to be loaded only once. It should help in the following problems:
- Performance: no need to load dynamic map every time user enters the search page view on SPA.
- Efficiency for Google Maps pricing change: less unnecessary calls for a dynamic map.
- Reaction to a [bug](https://issuetracker.google.com/issues/35821412) when removing Google Map instance

### DIFF
--- a/src/components/SearchMap/SearchMap.js
+++ b/src/components/SearchMap/SearchMap.js
@@ -297,31 +297,33 @@ export class SearchMapComponent extends Component {
 }
 
 SearchMapComponent.defaultProps = {
+  className: null,
+  rootClassName: null,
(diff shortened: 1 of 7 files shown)


src/components/SearchMap/SearchMap.js
```diff
@@ -240,6 +241,7 @@ export class SearchMapComponent extends Component {
     const {
       className,
       rootClassName,
+      reuseableContainerClassName,
```
lyyder: This could be declared in the component prop types definition.

PR #893, opened 2018-08-09
Generic location autocomplete

kpuputti:
This PR generalises the `LocationAutocompleteInput` so that the geocoding implementation can be swapped.

Two geocoder implementation are provided, one for Google Maps Places API (default) and one for Mapbox Geocoding API. The Mapbox geocoder requires `window.mapboxgl` and `window.mapboxSdk` globals to be present.


src/components/LocationAutocompleteInput/LocationAutocompleteInputImpl.js
```diff
@@ -1,11 +1,18 @@
...
 
-import css from './LocationAutocompleteInputGoogleMaps.css';
+import css from './LocationAutocompleteInput.css';
+
+// When using the Google Maps Places API for geocoding, a "Powered by
+// Google" logo should be shown next to the results. Turn this to
+// `false` when using some other Geocoding API. Autocomplete
+// predictions dropdown bottom padding might need adjusting as well to
+// hide the space left for the logo.
+const SHOW_POWERED_BY_GOOGLE = true;
```
Gnito: Would it be safer to use `config.googleMapsAPIKey` check somehow here? Especially if we end up making mapbox our default map provider? (Missing to turn this on/off is a clear violation of their terms.)

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ way to update this template, but currently, we follow a pattern:
 ---
 ## Upcoming version 2018-08-XX
 
+* [change] Removed the `country` parameter from the search page as it
+  was not used anywhere.
+  [#893](https://github.com/sharetribe/flex-template-web/pull/893)
+
```
lyyder: The changelog could contain even multiple entries for this PR. At least for adding the Mapbox geocoder and maybe also a note about the generalization of the geocoder.

src/components/LocationAutocompleteInput/LocationAutocompleteInputImpl.js
Gnito: Even better would be that the whole div (or just the flag) is exported from Geocoder file.

kpuputti: Could be better but a config var might make customisers think that it is enough to change the provider.

I think a config still makes sense, then we can already adjust the predictions list paddings etc. based on that. We just have to mention that some imports have to be changes too.

CHANGELOG.md
kpuputti: Is it ok not to advertise it yet? It's not usable without the script imports, and it would be a violation of the service terms to use the Mapbox geocoder without also using the Mapbox map (which isn't finished). We'll have a big major release when all the parts are ready.

src/components/LocationAutocompleteInput/LocationAutocompleteInputImpl.js
kpuputti: I moved the attribution element to the geocoder.

However, the predictions list leaves bottom padding for the attribution, and that padding is added in many places. I added a global var that can be used to set the padding when the map provider is changed.

src/forms/EditListingLocationForm/EditListingLocationForm.css
```diff
@@ -50,7 +50,7 @@
 .predictionsRoot {
   position: absolute;
   width: 100%;
-  padding-bottom: 98px;
+  padding-bottom: var(--locationAutocompleteBottomPadding);
```
kpuputti: I think the larger padding here was for the Mobile Safari bottom spacing, but it looked fine with 72px when I tested it.

CHANGELOG.md
lyyder: Valid point. Let's add an entry when Mapbox can properly be taken into use.

PR #895, opened 2018-08-10
Autocomplete current location

kpuputti:
This PR adds support for adding a "Current location" as a default search in the `LocationAutocompleteInput`. The location is queried from the browser geolocation API, and a bounding box is expanded around the given location.

~~Currently this is only implemented for the Google Maps geocoder as the bounding box calculation uses helpers from the Google Maps API.~~ (edit: also implemented in the Mapbox geocoder now)

## Screenshots

screen shot 2018-08-13 at 11 18 57

current-location-with-icon


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ way to update this template, but currently, we follow a pattern:
 ---
 ## Upcoming version 2018-08-XX
 
+* [change] Add support for user's current location as a default
+  suggestion in the location autocomplete search.
+  [#895](https://github.com/sharetribe/flex-template-web/pull/895)
```
lyyder: BTW, we have also agreed on using `add` entry tag in changelog.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ way to update this template, but currently, we follow a pattern:
 ---
 ## Upcoming version 2018-08-XX
 
+* [change] Add support for user's current location as a default
```
lyyder: BTW, we have also agreed on using `add` entry tag in changelog.

src/components/LocationAutocompleteInput/GeocoderGoogleMaps.js
```diff
@@ -1,8 +1,11 @@
 import React from 'react';
-import { getPlacePredictions, getPlaceDetails } from '../../util/googleMaps';
-
+import { getPlacePredictions, getPlaceDetails, locationBounds } from '../../util/googleMaps';
+import { userLocation } from '../../util/maps';
 import css from './LocationAutocompleteInput.css';
 
+export const CURRENT_LOCATION_ID = 'current-location';
+const CURRENT_LOCATION_BOUNDS_DISTANCE = 1000; // meters
```
lyyder: Should this be placed in config.js? Then marketplace customizer could easily modify it.

kpuputti: I think we have to think about the whole maps config when we can consider the Mapbox support in its entirety. I'd keep these a bit hidden for now, especially since part of configuring the default searches would still be in the geocoder file. Splitting that into several places wouldn't be any better.

We already have some Google Maps specific options in the config. We should maybe have separate keys for Google Maps and Mapbox that contain all the config for those map providers. So let's not hurry now and add random partial setting to the global config...

How does this sound?

lyyder: Sounds good, let's see how it looks when Mapbox is in place.

PR #896, opened 2018-08-13
Mapbox Map component

kpuputti:
This page implements a version of the `Map` component (used in `ListingPage`) using Mapbox.

### Static map with marker

screen shot 2018-08-14 at 15 22 42

### Dynamic map with marker

screen shot 2018-08-14 at 15 23 03

### Static map with custom marker

screen shot 2018-08-14 at 15 23 49

### Dynamic map with custom marker

screen shot 2018-08-14 at 15 24 01

### Static map with fuzzy location circle

screen shot 2018-08-14 at 15 25 04

### Dynamic map with fuzzy location circle

screen shot 2018-08-14 at 15 25 14


src/components/Map/Map.js
```diff
@@ -3,7 +3,8 @@ import { bool, number, object, string } from 'prop-types';
 import classNames from 'classnames';
 import { propTypes } from '../../util/types';
 import config from '../../config';
-import { StaticMap, DynamicMap, isMapsLibLoaded } from './GoogleMap';
+// import { StaticMap, DynamicMap, isMapsLibLoaded } from './GoogleMap';
```
lyyder: Is this comment something we'd leave to the code as an example on how to change the map implementation. Or remove when the map migration is complete?

src/components/Map/Map.js
```diff
@@ -45,8 +45,10 @@ export class Map extends Component {
       />
     ) : (
       <DynamicMap
-        containerElement={<div className={classes} onClick={this.onMapClicked} />}
```
lyyder: Where has the `onClick` gone?

kpuputti: It wasn't used at all, there wasn't even a `onMapClicked` method that it was calling. It was probably copy pasted from `SearchMap` that had such method.

src/components/Map/StaticMapboxMap.js
```diff
@@ -0,0 +1,90 @@
...
+  if (coordinatesConfig.fuzzy) {
+    return fuzzyCircleOverlay(center, coordinatesConfig);
+  }
+  if (coordinatesConfig.customMarker) {
+    return customMarkerOverlay(center, coordinatesConfig);
+  }
+  return markerOverlay(center);
+};
+
+class StaticMapboxMap extends Component {
```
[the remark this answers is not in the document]
kpuputti: Yes, I can change to a function. But always creating a class also has the upside of not having to indent the whole component (and generating lots of diff) when it required state of lifecycle methods.

src/components/Map/Map.js
kpuputti: Oops, the Mapbox one should be commented out and the Google one in use. I'll fix that.

I'd leave the Mapbox imports as comments so we can easily test those while doing the whole map provider migration. We can decide to drop them when putting it all together in the end.

PR #900, opened 2018-08-16
Reorganize maps configuration

kpuputti:
**BREAKING CHANGE**

This PR reorganises the maps configuration to work with both Google Maps and Mapbox. The structure of the configuration is different and paves the way for the bigger change of making Mapbox as the default map provider.

- [ ] Reorganise maps config
- [ ] Document the config
- [ ] Update docs
- [ ] Test all possible options


src/config.js
```diff
@@ -262,57 +263,75 @@ const siteFacebookPage = 'https://www.facebook.com/Sharetribe/';
...
-  // proper value for the offset and the radius, see:
-  //
-  // https://gis.stackexchange.com/a/8674
-
-  // Amount of maximum offset in meters that is applied to obfuscate
-  // the original coordinates. The actual value is random, but the
-  // obfuscated coordinates are withing a circle that has the same
-  // radius as the offset.
-  coordinateOffset: 500,
+const maps = {
```
lyyder: We could call this `map` as the application is using one map at a time. But I guess `maps` makes sense too.

kpuputti: Well, you sort of configure all the maps in the app. And nothing prevents using two or more maps at a time. But mainly `maps` is a bit more greppable than `map`.

PR #901, opened 2018-08-16
Availability improvements

lyyder:
Fixes two problems with availability:

- Makes a second time slots request if the number of bookable day exceeds 90 which is the number of time slots that can be fetched with one request
- Drops date time information form time slots request

### DIFF
--- a/src/containers/ListingPage/ListingPage.duck.js
+++ b/src/containers/ListingPage/ListingPage.duck.js
@@ -203,8 +203,13 @@ export const fetchTimeSlots = listingId => (dispatch, getState, sdk) => {
   const bookingRange = config.dayCountAvailableForBooking - 1;
   const timeSlotsRange = Math.min(bookingRange, maxTimeSlots);
 
-  const start = moment().toDate();
+  const start = moment
+    .utc()
+    .startOf('day')
+    .toDate();
   const end = moment()
+    .utc()
+    .startOf('day')
     .add(timeSlotsRange, 'days')
     .toDate();
   const params = { listingId, start, end };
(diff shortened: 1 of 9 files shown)


src/containers/ListingPage/ListingPage.duck.js
```diff
@@ -185,13 +185,56 @@ export const fetchReviews = listingId => (dispatch, getState, sdk) => {
...
+          start: end,
+          end: moment(end)
+            .add(secondRange, 'days')
+            .toDate(),
+        };
+
+        dispatch(timeSlotsRequest(secondParams)).then(secondBatch => {
+          const combined = timeSlots.concat(secondBatch);
+          dispatch(fetchTimeSlotsSuccess(combined));
+        });
```
kpuputti: Should there be a `return` before the `dispatch` or a `.catch` on the promise to handle error in the request?

lyyder: The `.catch` is left out so that in case of an error the `.catch` of the outer promise handles it. This way there error action is fired in case either of the time slot requests fail. A succeed action is dispatched when both requests succeed.

lyyder: How would a `return` help with handling a failing request here?

kpuputti: The failing promise never goes to the outer `.catch` without the `return`.

lyyder: Well so it seems. I was sure I had this tested. Well spotted 👍

lyyder:
Added the missing `return` statement. Requesting time slots on demand might introduce latency in showing the time slots so decided to load the time slots in `loadData`.

PR #905, opened 2018-08-21
Add path check for search page

lyyder:
When using `ReusableMapContainer` map components `onIdle` function can be invoked on other pages than `SearchPage` in case the window is resized. This PR adds a check that the current location is the search page before conducting a new search in `onIdle`.


src/containers/SearchPage/SearchPage.js
```diff
@@ -116,10 +116,17 @@ export class SearchPageComponent extends Component {
     const viewportBoundsChanged =
       this.viewportBounds && !hasSameSDKBounds(this.viewportBounds, viewportBounds);
 
+    const routes = routeConfiguration();
+    const searchPagePath = pathByRouteName('SearchPage', routes);
+    const currentPath = window && window.location && window.location.pathname;
```
kpuputti: Could this be done by passing down the `currentPage="SearchPage"` and reading that prop? I think that is done elsewhere in similar cases and would drop the need to read `window` or the URL.

lyyder: @kpuputti how would this work in practice? Wouldn't the `currentPage` always be set to `"SearchPage"` in the scope of the `onIdle` function?

lyyder: I think the `currentPage` prop has been used with Topbar which is introduced separately to each page.

kpuputti: Argh, you are right, there's only the single place in SearchPage that the map is rendered. Never mind then.

One more thing: if you add the `window` check, the correct way that doesn't  break server rendering is `typeof window !== 'undefined'`, although this function shouldn't be called in the server anyways.

PR #906, opened 2018-08-22
Disable default predictions in listing editing wizard

lyyder:
Adds a prop to `LocationAutocompleteInputImpl` that can be passed to `LocationAutocompleteInput` to disable default predictions in a location input.


lyyder:
One option here would also be that the default predictions list could be passed as prop and it would override the one from the geocoder. This way for example only current location could be used in the listing wizard. Downside in that would be that the geocoder specific prediction schema would spread around.

kpuputti:
The default prediction configuration will change in #900 and will be independent of the geocoder in use. However, current location cannot really be used in the listing wizard as it only returns coordinates, not address or bounds. So this solution is fine IMO.

PR #908, opened 2018-08-24
Fix mapbox reattachement

Gnito:
Check commit comments.

### DIFF
--- a/src/components/SearchMap/SearchMap.js
+++ b/src/components/SearchMap/SearchMap.js
@@ -43,8 +43,12 @@ export class SearchMapComponent extends Component {
 
     this.listings = [];
     this.mapRef = null;
-    this.mapReattachmentCount = 0;
-    this.state = { infoCardOpen: null };
+
+    if (typeof window !== 'undefined' && !window.mapReattachmentCount) {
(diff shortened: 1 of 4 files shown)


src/components/SearchMap/SearchMap.js
```diff
@@ -43,8 +43,12 @@ export class SearchMapComponent extends Component {
 
     this.listings = [];
     this.mapRef = null;
-    this.mapReattachmentCount = 0;
-    this.state = { infoCardOpen: null };
+
+    if (typeof window !== 'undefined' && !window.mapReattachmentCount) {
+      window.mapReattachmentCount = 0;
+    }
+    const mapReattachmentCount = window.mapReattachmentCount || 0;
```
lyyder: Is there a need here of the `|| 0`? Or what would be a case where the `window.mapReattachmentCount` would not be set to `0` on line 48 and then here the `mapReattachmentCount` would evaluate to `0` because the `window.mapReattachmentCount` is not available?

Gnito: @lyyder server-side rendering (since this is in a constructor).

PR #910, opened 2018-08-24
Fix search after navigation

lyyder:
Mapbox does not update when performing a search with the following steps:
 - Search for "Helsinki"
 - Navigate away from search page, to for example a listing page
 - Search for "New York" in top bar

### DIFF
--- a/src/components/SearchMap/SearchMapWithMapbox.js
+++ b/src/components/SearchMap/SearchMapWithMapbox.js
@@ -241,10 +241,10 @@ class SearchMapWithMapbox extends Component {
   componentDidUpdate(prevProps) {
     if (!this.map) {
       this.initializeMap();
-
-      /* Notify parent component that Mapbox map is loaded */
-      this.props.onMapLoad(this.map);
     }
+
+    /* Notify parent component that Mapbox map is loaded */
+    this.props.onMapLoad(this.map);
   }
 
   onMount(element) {


lyyder:
@Gnito The 773023d commit changes the Mapbox implementation so that the onMapLoad is called always when the component updates. This change makes the map update but it still has some problems:

 - The map update invokes an `onIdle` call. So does the Google Maps implementation but for some reason only with using google maps the `SearchPage` component's [componentWillReceiveProps function is called which clears this.viewportbounds](https://github.com/sharetribe/flex-template-web/blob/fix-search-after-navigation/src/containers/SearchPage/SearchPage.js#L87)
 - As `this.viewportBounds` is not cleared [the viewportBoundsChanged check will evaluate to true](https://github.com/sharetribe/flex-template-web/blob/fix-search-after-navigation/src/containers/SearchPage/SearchPage.js#L109) and this will lead to a history push with the `mapSearch` param set `true`. Using Google Maps, the `viewportBoundsChanged` will evaluate to `null` and the history push won't be invoked in onIdle.
 - Why the `componentWillReceiveProps` on `SearchPage` is not invoked when using Mapbox is a bit of a mystery

Gnito:
Thanks for the initial input - it was useful comment even though I ended up fixing this in another branch. 

I'm closing this branch.

PR #930, opened 2018-09-14
Fix obfuscate coordinates implementation

Gnito:
Some coordinates were outside of given offset radius.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ way to update this template, but currently, we follow a pattern:
 
 ---
 ## Upcoming version 2018-09-XX
+* [fix] obfuscatedCoordinatesImpl didn't always return coordinates within given
+  offset radius.
+  [#930](https://github.com/sharetribe/flex-template-web/pull/930)
 * [fix] LocationAutocompleteInput: blur input when selecting by enter
   to prevent flash of default predictions.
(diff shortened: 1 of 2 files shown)


Gnito:
Post review more than welcome

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ way to update this template, but currently, we follow a pattern:
 
 ---
 ## Upcoming version 2018-09-XX
+* [fix] obfuscatedCoordinatesImpl didn't always return coordinates within given
+  offset radius.
```
kpuputti: Just a generic note on the abstraction level of this description: I think it is too detailed. Something like "Improved fuzzy location calculation" would probably suffice for people who read the changelog.

src/util/maps.js
```diff
@@ -5,41 +5,67 @@ import config from '../config';
...
+  return { lat, lng };
+};
+
 /**
  * This obfuscatedCoordinatesImpl function is a temporary solution for the coordinate obfuscation.
  * In the future, improved version needs to have protectedData working and
  * available in accepted transaction.
+ *
+ * Based on:
+ * https://gis.stackexchange.com/questions/25877/generating-random-locations-nearby#answer-213898
```
kpuputti: I truly wish we had better sources for reference than a random answer with three upvotes :/ But I hope you tested this well, no need to change if it works.

PR #939, opened 2018-10-15
Redux doc

Gnito:
We got feedback that Redux/duck files need some documentation.

### DIFF
--- a/docs/redux.md
+++ b/docs/redux.md
@@ -1,0 +1,126 @@
+# Redux configuration
+
+## What is Redux
+
+Flex Template for Web (FTW) is JavaScript single-page application (SPA). This means that the app
+needs to be able to render several different layouts (pages) depending on user interaction. State
+management is essential for this process. FTW needs to know if a user has been authenticated, if it
+has received relevant data for the current page, and so on.
+
+We use [Redux](https://redux.js.org/introduction) for state management on the application level. You
+should read more about Redux before you start modifying queries to Flex API or creating new Page
+level elements (unless you are modifying [a static page](static-pages.md)).
+
+In the following subtopics, we assume that you know the
+[basics of Redux](https://redux.js.org/basics) already.
+
+* [Containers: Pages + TopbarContainer](#containers-pages--topbarcontainer)
+* [Duck files](#duck-files)
+* [Setting up Redux](#setting-up-redux)
+* [Advanced Redux concepts: thunks](#advanced-redux-concepts-thunks)
+
+## Containers: Pages + TopbarContainer
+
+We have set up FTW so that pages are aware of Redux state store, but other components and forms are
+purely
+[presentational components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0).
+This makes it easier to customize UI components - you don't need to be aware of the complexity
+related to Redux setup when you just want to touch the behavior of some visual component. In those
+cases, you can just head to `scr/components/` or `src/forms/` directory and you can see from there
+what props are available for each component when they are rendered.
+
+Naturally, there is a need for higher level components which fetch new data and define what props
+are passed down to presentational components. In Redux terminology, those components are called
+_Containers_. FTW has defined all the containers inside a directory called `src/containers/`. It
+contains all the pages and a special container for top bar (TopbarContainer) since that is a very
+complex component and it is shared with almost every page. You can read more about differences
+between presentational and container components from an
+[article written by Dan Abramov](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)
+
+The actual container setup of a page level component can be found from the bottom of the component
+file. For example, `src/containers/TransactionPage/TransactionPage.js` connects itself to Redux
+store with `mapStateToProps` and `mapDispatchToProps` functions:
+
+```js
+const TransactionPage = compose(connect(mapStateToProps, mapDispatchToProps), injectIntl)(
+  TransactionPageComponent
+);
+```
+
+## Duck files
+
+Inside `src/containers/<ComponentName>` directory, we have also a special `<ComponentName>.duck.js`
+file. These files wrap all the other page-specific Redux concepts into a single file. Instead of
+writing action types, action creators and reducers in separate files (and spread them around our
+directory structure), FTW tries to keep pages encapsulated. Page specific actions, store updates,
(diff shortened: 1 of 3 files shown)


docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+* [Duck files](#duck-files)
+* [Setting up Redux](#setting-up-redux)
+* [Advanced Redux concepts: thunks](#advanced-redux-concepts-thunks)
+
+## Containers: Pages + TopbarContainer
+
+We have set up FTW so that pages are aware of Redux, but other components and forms are purely
+presentational components. This makes it easier to customize UI components - you don't need to be
+aware of the complexity related to Redux setup when you just want to touch the behavior of some
+visual component. In those cases, you can just head to `scr/components/` or `src/forms/` folder and
```
kpuputti: I would always use "directory" instead of "folder":

https://stackoverflow.com/questions/5078676/what-is-the-difference-between-a-directory-and-a-folder

docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+
+* [Containers: Pages + TopbarContainer](#containers-pages--topbarcontainer)
+* [Duck files](#duck-files)
+* [Setting up Redux](#setting-up-redux)
+* [Advanced Redux concepts: thunks](#advanced-redux-concepts-thunks)
+
+## Containers: Pages + TopbarContainer
+
+We have set up FTW so that pages are aware of Redux, but other components and forms are purely
+presentational components. This makes it easier to customize UI components - you don't need to be
```
kpuputti: The "container" and "presentational" terms are not defined anywhere here and might be confusing. I don't think even the Redux basics docs mention them. Maybe link to Abramov's Medium post about them or add a simple explanation here. Those are quite crucial concepts to grok.

docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+aware of the complexity related to Redux setup when you just want to touch the behavior of some
+visual component. In those cases, you can just head to `scr/components/` or `src/forms/` folder and
+you can see from there what props are available for each component when they are rendered.
+
+Naturally, there is a need for higher level components which fetch new data and define what props
+are passed down to presentational components. In Redux terminology, those components are called
+_Containers_. FTW has defined all the containers inside a folder called `src/containers/`. It
```
kpuputti: Ok, here is part of the explanation for "containers". Maybe add something earlier for both "presentational" and "containers"?

docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+contains all the pages and a special container for top bar (TopbarContainer) since that is a very
+complex component and it is shared with almost every page.
+
+The actual container setup of a page level component can be found from the bottom of the component
+file. For example, `src/containers/TransactionPage/TransactionPage.js` connects itself to Redux
```
kpuputti: If all there dir/file refs would be links, that would help a lot!

docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Duck files
+
+Inside `src/containers/<ComponentName>` folder, we have also a special `<ComponentName>.duck.js`
+file. These files wrap all the other page-specific Redux concepts into a single file. Instead of
+writing action types, action creators and reducers in separate files (and spread them around our
+folder structure), FTW tries to keep pages isolated. Page specific actions, store updates, and data
+fetches happen inside their respective folder - just look for a file which name follows a pattern:
+`<ComponentName>.duck.js`. This pattern was first proposed by Erik Rasmussen and you can read more
+about it from [his repository](https://github.com/erikras/ducks-modular-redux).
```
kpuputti: Some suggestions:

- folder -> directory
- isolated -> encapsulated

Maybe add a sentence that ducks are really a simple concept of just adding related things into a single file.

docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+`appReducer` is called by `createReducer` function, which is called inside of `configureStore`
+function (in `src/store.js`).
+
+This setup creates a store structure that separates container specific state as well as global data
+by their reducer names. Together with Ducks module naming schema, this means that:
+
+* the state of the `ListingPage` can be found from `state.ListingPage` and
+* the state of the global `user` object can be found from `state.user`.
+
+## Advanced Redux concepts: thunks
```
kpuputti: Not really "advanced", but ok. :P

docs/redux.md
```diff
@@ -0,0 +1,118 @@
...
+One essential part of state management in FTW, is filling the Redux store with data fetched from the
+Flex API. This is done with
+[Redux Thunks](https://redux.js.org/advanced/asyncactions#async-action-creators), which is a Redux
+middleware to create asynchronous action creators.
+
+As with every other Redux store actions, we have defined Thunks inside `*.duck.js` file. For
+example, fetching listing reviews can be done with a following thunk function:
+
+```js
+export const fetchReviews = listingId => (dispatch, getState, sdk) => {
```
kpuputti: Should there be a note about Redux middleware? i.e. where the `sdk` comes from?

docs/redux.md
Gnito: I added a link to the first occurrence of "presentational component" and another "read more" pointer to the end of this paragraph.

docs/redux.md
Gnito: Changed. (even though Rasmussen uses "isolated")

docs/redux.md
Gnito: It is advanced since Redux docs say it's advanced :P

docs/redux.md
Gnito: Good point. I added this to the end of that example:

> Note: `sdk` parameter is also provided by Redux Thunk. We pass it to middleware in store.js: `thunk.withExtraArgument(sdk)`

Gnito:
> Some interesting things that could/should be there:
>
> * How pages load their data (loadData, routing, ducks)
> * How data from API is normalized and denormalized for the store and for components
> * How incomplete data is rendered and should be wrapped with ensureX functions
> * Where to draw the line between containers and presentational components (i.e. which components own the data)

I think these need their own document or at least bigger section that just didn't fit my document-something-on-Monday routine.

- Routing + loadData are pretty tight together (need refs to each others)
- Normalization/denormalization (+ maybe writing about ensureX on the same go)

I think that writing something deeper about container vs presentational components is something that needs to be learnt from better sources than this repo. Our take is already mentioned here (i.e. pages + TopbarContainer). Of course, this document could be improved over time.

PR #940, opened 2018-10-16
Referrer policy updates

Gnito:
This update is to ensure that page with URLs that contain tokens from API, doesn't leak that information to external links. (Including links in `<head>` section, CSS files etc.)

> NOTE: remember to use `<ExternalLink>` component always when dealing with external links.
> [ExternalLink](https://github.com/sharetribe/flex-template-web/blob/master/src/components/ExternalLink/ExternalLink.js)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ way to update this template, but currently, we follow a pattern:
 ---
 
 ## Upcoming version 2018-10-XX
-
+* [add] Add referrer policy due tokens in URL on PasswordResetPage and EmailVerificationPage.
+  [#940](https://github.com/sharetribe/flex-template-web/pull/940)
 * [add] Added initial documentation about our Redux setup.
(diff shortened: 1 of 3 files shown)


Gnito:
@bladealslayer @kpuputti 

FTW renders Helmet meta tags before we load anything external on public/index.html:
```html
<html data-htmlattr="htmlAttributes">
  <head>
    <meta charset="utf-8">
    <!--!title-->
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <!--!meta-->
```

I updated this PR so that referrer is actually handled in Page.js, where it can be set as the first meta tag (in Helmet handled meta tags) 

Furthermore, I had to modify also canonical URL creation so that it doesn't contain query & fragment sections when referrer is enforced. (Including params is necessary for SearchPage and other situations where URL params affect to the content of the page.)

This setup returns origin for the first external URL on `http://localhost:4000/reset-password?asdf=asdf`
screen shot 2018-10-17 at 12 49 35

PR #949, opened 2018-11-12
Date filter

OtterleyW:
**Note:**
This PR adds **date filter** and **refactors other filters** to be more consistent: there are now plain and popup versions of _select-single_ and _select-multiple_ filters too.

----

bookingdaterangefilterplain
**BookingDateRangeFilter on mobile version**
- Automatically apply selected dates
- Clear selection when "Clear" is clicked

bookingdaterangefilterpopup
**BookingDateRangeFilter on desktop version**
- Open in popup
- Select dates when "Apply" is clicked
- Cancel resets selection back to previous selected dates


Gnito:
Reorder commits so that "Add help functions for ISO8601 date format" is included before "Add DateRangeFilter to desktop and mobile versions". (i.e. functions are included before they are used)

src/util/dates.js
```diff
@@ -140,3 +140,11 @@ export const formatDate = (intl, todayString, d) => {
 
   return `${formattedDate}, ${formattedTime}`;
 };
+
```
lyyder: A comment for these ISO8601 functions would be nice. Like the ones on other helper functions in this file.

PR #955, opened 2018-11-19
Translations management CLI

lyyder:
translations_cli

Adds a command line tool that can be used to manage translations:

- see what keys are missing compared to English translations
- add a missing translation

__Testing__

To try the out the tool, copy `src/translations/en.json` to `src/translations/es.json` for example and remove a few lines from the es translations file.

### DIFF
--- a/translations
+++ b/translations
@@ -44,7 +44,6 @@ const run = () => {
       targetKeys = Object.keys(target);
 
       const diff = difference(sourceKeys, targetKeys);
-      console.log(`target: ${target}`);
 
       if (diff.length > 0) {
         return selectKey(targetLang, diff, source, target);
--- a/translations
+++ b/translations
@@ -56,6 +56,7 @@ const targetLangChoices = () => {
   const filenames = fs.readdirSync(folder);
 
   const choices = filenames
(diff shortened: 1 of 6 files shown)


lyyder:
Let's see at what point we get the default languages translation files added to the project. Not the tool expects to have _es_, _de_, and _fr_ translations in place. Maybe all the language definitions could be removed and added as the corresponding files are added. That way this could be merged now without waiting for the translations to finish.

translations
```diff
@@ -0,0 +1,183 @@
...
+  let sourceKeys;
+  let targetKeys;
+  let key;
+
+  selectLanguage()
+    .then(answers => {
+      sourceLang = SOURCE_LANG.value;
+      targetLang = answers.lang;
+      source = require(filePath(sourceLang));
+      target = require(filePath(targetLang));
```
Gnito: It's probably better to not use `require` to future-proof this since _"Require is synchronous function and is called only once, which means the calls receive a cached result. If the file is updated you can't re-read it using this method"_

https://stackabuse.com/reading-and-writing-json-files-with-node-js/

PR #960, opened 2018-11-22
Show Stripe error message if error contains Stripe error

OtterleyW:


src/util/errors.js
```diff
@@ -136,6 +136,23 @@ export const isTransactionInitiateAmountTooLowError = error => {
...
+/**
+ * Check if the given API error (from `sdk.transaction.initiate()`) is
+ * due to other error in Stripe.
+ */
+export const isTransactionInitiateStripeError = error => {
+  if (error) {
+    return errorAPIErrors(error).reduce((messages, apiError) => {
+      const isPaymentFailedError =
+        apiError.status === 402 && apiError.code === ERROR_CODE_PAYMENT_FAILED;
+      let hasMessage = apiError && apiError.meta && apiError.meta.stripeMessage;
```
Gnito: No need to use `let` since there's no need to update the variable inside its lifespan (i.e. this function block).

src/util/errors.js
```diff
@@ -136,6 +136,23 @@ export const isTransactionInitiateAmountTooLowError = error => {
...
+
+      return isPaymentFailedError && hasMessage ? [...messages, hasMessage] : messages;
```
Gnito: This could be written as:
```js
const stripeMessageMaybe = isPaymentFailedError && apiError && apiError.meta
  ? [apiError.meta.stripeMessage]
  : [];
return [...messages, ...stripeMessageMaybe];
```

src/util/errors.js
```diff
@@ -136,6 +136,23 @@ export const isTransactionInitiateAmountTooLowError = error => {
   });
 };
 
+/**
+ * Check if the given API error (from `sdk.transaction.initiate()`) is
+ * due to other error in Stripe.
+ */
+export const isTransactionInitiateStripeError = error => {
```
Gnito: Since this is not returning any boolean, the name should be something else. 
How about `transactionInitiateOrderStripeErrors` or `transactionInitiateStripeErrors`?

PR #967, opened 2018-11-29
Update dependencies

OtterleyW:
Updating the Prettier caused changes in multiple files:
- 22 .md files changed, this affects mostly to documentation files
- in ~26 files compose setup is split into multiple lines
- 12 other file changes were mostly about:
  - calculations got brackets for readability
  - ternaries got indentation changes

You can see the files changed in [Update Prettier to 1.15.3](https://github.com/sharetribe/flex-template-web/pull/967/commits/0087a78ba27249af75faecae26f84291c9fef8c8) commit

Also updating Enzyme and react-dates caused some file changes.


package.json
```diff
@@ -64,7 +64,7 @@
     "nodemon": "^1.17.2",
     "nsp": "^3.2.1",
     "nsp-preprocessor-yarn": "^1.0.1",
-    "prettier": "^1.11.1"
+    "prettier": "^1.15.3"
```
Gnito: This update brought up one annoying problem related to Prettier. 
updating it might affect quite many files, which might be a problem for our customers. 

So, we have few annoying options:
- **a)** We keep updating Prettier 
  * We could update case-by-case or
  * Tie Prettier update to major FTW release
-  **b)** We freeze Prettier updates to the current version
  * This means that if our clients update their version of Prettier, they will have conflicts if they ever try to update from upstream again.
- **c)** We remove Prettier (I don't think anybody wants to do this option)

----
This particular Prettier update had following consequences:
- 22 docs/.md files changed (this is probably not a problem to our customizers)
- ~26 compose setup is split to multiple lines <sup>1</sup> (These are mostly Reduc Containers. This is probably not a problem to our clients either.)
- 12 other file changes were mostly about:
  * calculations got brackets for readability<sup>2</sup> and 
  * ternaries got indentation changes<sup>3</sup>

--
[1]
```js
const CheckoutPage = compose(
  withRouter,
  connect(
    mapStateToProps,
    mapDispatchToProps
  ),
  injectIntl
)(CheckoutPageComponent);
``` 

[2]: e.g. `lat * pi / 180` -> `(lat * pi) / 180`.
[3]: [ternary indentation](https://prettier.io/blog/2018/11/07/1.15.0.html#flatten-else-branch-for-nested-ternaries-5039-by-suchipi-5272-by-duailibe-5333-by-ikatyang)

Gnito: My suggestion, for now, is that we include these changes, but add a documentation about how we are planning to update Prettier in the future.

Any comments?

lyyder: > if our clients update their version of Prettier, they will have conflicts

How does this work? Someone has installed prettier globally?

Anyhow, option __a)__ is probably the way to go. Just announce it in release notes accordingly.

Gnito: @lyyder would you prefer a1 (case-by-case) or a2 (only when changing major release)?

lyyder: Do you mean that would the Prettier update _always_ be tied to a major version release or would we check on every update that should it be part of a major version release or a more minor one?

I guess case-by-case. Does it make sense to create a rule to always tie the Prettier updates to major versions if the updates themselves are small?

PR #976, opened 2018-12-10
Enable requesting to book straight from enquiry

lyyder:
Show `BookingPanel` on tx page for enquired transactions and allow the user to create a booking straight from the enquiry. The transaction created for the enquiry is transitioned to create the booking so all the previous discussions related to that enquiry are available after the booking.

enquiry

When creating a booking after an enquiry, the _initial message_ input is not shown on checkout page:

checkout


src/routeConfiguration.js
```diff
@@ -193,7 +193,7 @@ const routeConfiguration = () => {
       auth: true,
       authPage: 'LoginPage',
       component: props => <TransactionPage {...props} transactionRole="customer" />,
-      loadData: TransactionPage.loadData,
+      loadData: params => TransactionPage.loadData({ ...params, transactionRole: 'customer' }),
```
Gnito: 👍

src/components/TransactionPanel/TransactionPanel.helpers.js
```diff
@@ -106,7 +106,7 @@ export const AddressLinkMaybe = props => {
   const fullAddress =
     typeof building === 'string' && building.length > 0 ? `${building}, ${address}` : address;
 
-  return (isProvider || txIsAcceptedForCustomer) && hrefToGoogleMaps ? (
+  return txIsAcceptedForCustomer && hrefToGoogleMaps ? (
```
Gnito: Hmm. A provider could see the location of his listing all the time, but this works too.

lyyder: @Gnito I don't quite follow. The idea is to hide the address link from the provider.

lyyder:
@Gnito would the time slots be cached in the marketplace data? A problem with that is that the available slots might not be up to date when viewing availability of a listing. But some sort of common solution that would reduce the duplication would be welcome indeed.

src/components/TransactionPanel/TransactionPanel.helpers.js
Gnito: @lyyder provider owns the listing - he knows where it is.

Gnito:
Yeah, that solution "draft" would still need to be updated when a user goes to a page that uses timeslots. Anyway, this is just a start for a possible solution - it can be solved later.

src/components/TransactionPanel/TransactionPanel.helpers.js
lyyder: Yes, and that's why it's not relevant to show the address link to the provider.

lyyder: Solved IRL: When to show the listing location link has been a topic of a debate also before and probably continues to be one in the future. For now, we'll only show it only to customers after a booking has been accepted.

PR #978, opened 2018-12-17
NSP is dead and the API no longer resolves - replace with yarn audit.…

rdoh:
… see: https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,7 +91,7 @@ jobs:
             - node_modules/
       - run:
           name: Audit dependencies
-          command: yarn run audit
+          command: yarn audit
 workflows:
   version: 2
   format_test_build_audit:
--- a/package.json
+++ b/package.json
@@ -78,8 +78,7 @@
     "eject": "sharetribe-scripts eject",
     "start": "node server/index.js",
     "dev-server": "export NODE_ENV=development PORT=4000 REACT_APP_CANONICAL_ROOT_URL=http://localhost:4000&&yarn run build&&nodemon --watch server server/index.js",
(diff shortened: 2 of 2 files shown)


Gnito:
@rdoh thanks for the pull request, we were out of reach last week and haven't had time to fix this. We appreciate that there're pull requests or issues created out of deploy problems to alert others who have dependencies to this codebase.

That being said, we'll make another PR to handle the required changes a bit faster this time.

----

 I actually think we should drop the CI task altogether - until we have something better figured out. 

The problem is that there are lots of dependencies that come from Create React App 1.x.x version (like Prototype pollution with paths pointing to CRA fork: `sharetribe-scripts > jest > jest-cli > jest-runtime > jest-haste-map > sane > exec-sh > merge`), and we have exceptions for those in `.nsprc` file. 

So, in addition to changes mentioned in this PR, I'd propose that we:
1) Drop CI
2) Remove dev dependencies:
```
    "nsp": "^3.2.1",
    "nsp-preprocessor-yarn": "^1.0.1",
```
3) Keep package.json script `yarn run audit`, but just run `yarn audit` with it. (We would keep this mainly to prepare for future audit script and not to break existing work flow.)

TODO:
Now: remove circleci task
Future task: add audit script that checks against saved snapshot or exception list.

(Although, running `yarn audit --json` gives data worth of 10 MB - so, snapshot is probably not a good idea.)

OtterleyW:
Closing this PR because [#979](https://github.com/sharetribe/flex-template-web/pull/979) is now merged.

PR #980, opened 2018-12-17
Add stripe business accounts

OtterleyW:
`PayoutDetailsForm` is now divided to multiple smaller subcomponents:
- `PayoutDetailsCompany` and  `PayoutDetailsIndividual` contain needed fields for individual and company accounts
- `PayoutDetailsAddress` ,`PayoutDetailsBankDetails` and `PayoutDetailsPersonalDetails` specific fields related to address, bank details and personal details

New `FieldArray` component is created for adding multiple additional owners if needed.

## Select individual or company account
fieldradiobutton

## Add additional owners
additionalowners3


src/components/FieldArray/FieldArray.css
```diff
@@ -0,0 +1,19 @@
+@import '../../marketplace.css';
+
+.root {
+}
+
+.formRow {
```
Gnito: This is not used.

src/components/FieldArray/FieldArray.css
```diff
@@ -0,0 +1,19 @@
...
+  display: flex;
+  justify-content: space-between;
+  flex-shrink: 0;
+  width: 100%;
+  margin-bottom: 24px;
+}
+
+.field {
```
Gnito: And this is not used either.

src/components/FieldArray/FieldArray.css
```diff
@@ -0,0 +1,19 @@
...
+  width: 100%;
+}
+.buttonsWrapper button {
```
Gnito: This is used in the example only. (We try to keep example/styleguide CSS  out of main component CSS by having separate ComponentName.example.css file). 

Another thing to note: we don't want to refer to elements anywhere except in marketplaceIndex.css file - add classes instead.

src/components/FieldArray/FieldArray.js
```diff
@@ -0,0 +1,36 @@
...
+
+import css from './FieldArray.css';
+
+const FieldArrayComponent = props => {
+  const { rootClassName, className, id, name, children } = props;
+
+  const classes = classNames(rootClassName || css.root, className);
+
+  return (
+    <FieldArray id={id} name={name} className={classes}>
```
Gnito: Sorry, but I don't think it makes much sense for this to exists as a separate component (FieldArray wrapper for FieldArray). There are no styles attached to this component (and the styles are likely to be different in different use scenarios when `import { FieldArray } from 'react-final-form-arrays';` is used. 

It is also good to note that every time FieldArrays are used inside Final Form, the form needs to pass `arrayMutators` as props to the FinalForm component itself. So, wrapping the usage of `react-final-form-arrays` library with separate component like this one might be a bit confusing since it doesn't work out of the box.

src/components/IconClose/IconClose.js
```diff
@@ -5,14 +5,17 @@ import classNames from 'classnames';
 import css from './IconClose.css';
 
 const IconClose = props => {
-  const { className, rootClassName } = props;
+  const { className, rootClassName, size } = props;
   const classes = classNames(rootClassName || css.root, className);
 
+  // Use given size as width and height or default to 12px
+  const svgSize = size ? size : 12;
```
Gnito: There's a bit of catch when sizing SVG icons with width and height. They render graphics (paths, rects, circles etc.) to their own inner viewbox (or should I say coordinate system). 

There's one problem with this concept related to line graphics: if the width of the icon is scaled down, also the line width is scaled down accordingly. So, it is unlikely that the width of the line matches with the width our designer intended it to be. (This is the reason why we have different SVG graphics for different sizes in [IconArrowHead.js](https://github.com/sharetribe/flex-template-web/blob/master/src/components/IconArrowHead/IconArrowHead.js)

src/ducks/user.duck.js
```diff
@@ -389,28 +389,66 @@ export const createStripeAccount = payoutDetails => (dispatch, getState, sdk) =>
...
   dispatch(stripeAccountCreateRequest());
 
+  const { accountType, country } = payoutDetails;
+
+  let payoutDetailValues;
+  if (accountType === 'company') {
+    payoutDetailValues = payoutDetails['company'];
+  } else if (accountType === 'individual') {
+    payoutDetailValues = payoutDetails['individual'];
+  }
```
Gnito: This is OK but can be achieved with ternary too.

src/forms/PayoutDetailsForm/PayoutDetailsForm.js
```diff
@@ -56,49 +45,14 @@ const PayoutDetailsFormComponent = props => (
...
-      const firstNameLabel = intl.formatMessage({ id: 'PayoutDetailsForm.firstNameLabel' });
-      const firstNamePlaceholder = intl.formatMessage({
-        id: 'PayoutDetailsForm.firstNamePlaceholder',
-      });
-      const firstNameRequired = validators.required(
-        intl.formatMessage({
-          id: 'PayoutDetailsForm.firstNameRequired',
-        })
-      );
+      const { country } = values;
```
Gnito: You could extract accountType here too since it is used later.

src/forms/PayoutDetailsForm/PayoutDetailsForm.js
```diff
@@ -185,130 +96,87 @@ const PayoutDetailsFormComponent = props => (
...
-                disabled={disabled}
-                className={css.personalIdNumber}
-                type="text"
-                label={personalIdNumberLabel}
-                placeholder={personalIdNumberPlaceholder}
-                validate={personalIdNumberValid}
-              />
-            </div>
+          {values.accountType ? (
+            <React.Fragment>
```
Gnito: 👍

src/forms/PayoutDetailsForm/PayoutDetailsFormCompany.js
```diff
@@ -0,0 +1,198 @@
...
+    stripeCountryConfigs(country).companyConfig.personalAddress;
+
+  const showAdditionalOwnersField =
+    country &&
+    stripeCountryConfigs(country).companyConfig &&
+    stripeCountryConfigs(country).companyConfig.additionalOwners;
+
+  const hasAdditionalOwners = values.company && values.company.additionalOwners;
+  const hasMaxNumberOfAdditionalOwners =
+    hasAdditionalOwners && values.company.additionalOwners.length >= 4;
```
Gnito: We don't want to use magic numbers in the code. So, I'd suggest that this number for is either 
1) coming from config: `maxNumberOfAdditionalOwners` (might be unnecessary if the value is just 4 or nothing) or
2) it is a constant defined at the beginning of this file:
```
// In EU, there can be a maximum of 4 additional owners
const MAX_NUMBER_OF_ADDITIONAL_OWNERS = 4;
```

PR #983, opened 2018-12-18
Delete default builtin mail templates

anttiSalmivaara:
Other mail templates than those related to transactions can now be edited in console, so they are removed from the repository. At the same time the context variables are updated to match console and README is updated.


ext/default-mail-templates/README.md
```diff
@@ -1,7 +1,7 @@
 # Mail templates
 
-This directory contains the default templates for all mails sent by default as result of activity in
-the marketplace.
+This directory contains the default templates for mails defined in the default transaction process.
+Templates for other notifications can be edited in Console.
```
ovan: We could stick to same terminology that is used here: https://flex-docs.sharetribe.com/email-templates.html. So built-in emails and transaction emails.

PR #988, opened 2018-12-27
Layout changes to listing and tx (enquiry) page

lyyder:
- 3 baselines between the listing title and the Price per night on TX page (enquiry), it was 4.
- 8 baselines between the Price per night and the “Start date”, it was 10.
- 8 baselines between the Date’s line and the Booking breakdown, it was 9.

### DIFF
--- a/src/components/BookingPanel/BookingPanel.js
+++ b/src/components/BookingPanel/BookingPanel.js
@@ -90,7 +90,7 @@ const BookingPanel = props => {
     : 'BookingPanel.perUnit';
 
   const classes = classNames(rootClassName || css.root, className);
-  const titleClasses = classNames(css.bookingTitle, titleClassName);
+  const titleClasses = classNames(titleClassName || css.bookingTitle);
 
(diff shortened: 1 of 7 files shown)


src/components/BookingPanel/BookingPanel.js
```diff
@@ -89,6 +90,7 @@ const BookingPanel = props => {
     : 'BookingPanel.perUnit';
 
   const classes = classNames(rootClassName || css.root, className);
+  const titleClasses = classNames(css.bookingTitle, titleClassName);
```
Gnito: This just creates a string "BookingPanel__bookingTitle_asdfasd TransactionPage__bookingTitle_werwee"
Which means that both styles come from main.bundle.css and the import order might change the overriding rules.

maybe  change this to`classNames(titleClassName || css.bookingTitle)`

Gnito: And override all the values

PR #994, opened 2019-01-08
Config CLI

OtterleyW:
CLI for creating .env file and configuring required environment variables

config-cli


docs/translations.md
```diff
@@ -179,7 +179,7 @@ In case you have added a new language translation file and are pulling translati
 in sync. Running the following command in the project root
 
 ```
-node translations.js
+node scripts/translations.js
```
Gnito: Actually, it would make sense to create package.json script out of these `yarn run translations` or maybe even `yarn run translate`.

scripts/config.js
```diff
@@ -0,0 +1,401 @@
...
+ *
+ * @param {array} answers array of answers user has given to questions
+ * @param {string} line line from the existing .env file
+ *
+ * @returns key if the key matches to the one in the line
+ */
+const checkIfSameLine = (answers, line) => {
+  let foundKey;
+  if (answers) {
+    Object.keys(answers).map(key => {
```
lyyder: Instead of `map` maybe [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) could suit here better.

Another observation is that in the function docstring the type of `answers` is `array` but here `Object.keys` is invoked on it which indicates that it would be a normal `Object`. A sidenote here that in JavaScript array is a kind of object that provides some list-like functionalities and uses indices as keys. So calling `Object.keys` on an array is totally fine but it just returns the indices of the elements in the array:
```js
Object.keys(["foo", "bar", "xyz"]); // => [0, 1, 2]
```

scripts/config.js
```diff
@@ -0,0 +1,401 @@
...
+
+/**
+ * Check if line contains the key from the user answers.
+ *
+ * @param {array} answers array of answers user has given to questions
+ * @param {string} line line from the existing .env file
+ *
+ * @returns key if the key matches to the one in the line
+ */
+const checkIfSameLine = (answers, line) => {
```
lyyder: Could a more descriptive name make it easier to quickly see what this function does? Even something verbose like `getKeyFromAnswers` or something of the sort could maybe provide a bit of context when seeing this used in the code.

lyyder:
One more thing! Did you check this library?
https://www.npmjs.com/package/dotenv

Don't know if it provides any extra value but maybe worth checking if it could streamline reading the current `.env` file.

scripts/config.js
```diff
@@ -0,0 +1,401 @@
...
+ * @return array containing all keys and values saved to .env file
+ */
+const findSavedValues = () => {
+  const savedEnvFile = fs.readFileSync('./.env').toString();
+
+  const settings = savedEnvFile.split('\n').reduce((obj, line) => {
+    const splits = line.split('=');
+    const key = splits[0];
+    if (splits.length > 1) {
+      obj[key] = splits.slice(1).join('=');
```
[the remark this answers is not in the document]
OtterleyW: Yes (though I don't know if that is needed actually)

scripts/config.js
OtterleyW: Indeed there was a mistake in the docstring and `answers` actually is `Object` and not `array` otherwise I think Object.keys wouldn't have worked in this case 🤔

scripts/config.js
```diff
@@ -0,0 +1,401 @@
...
+ *
+ * @param {array} settings - array of values used as default if user is editing the .env file
+ *
+ * @returns array of questions passed to Inquirer
+ *
+ */
+const mandatoryVariables = settings => {
+  const clientIdDefault = settings
+    ? { default: settings.REACT_APP_SHARETRIBE_SDK_CLIENT_ID }
+    : undefined;
```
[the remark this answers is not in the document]
OtterleyW: I thought it was needed for object spread to work but since null works too I think using null is better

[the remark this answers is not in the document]
Gnito: @OtterleyW use empty object, then it's safe to spread.

```js
const clientIdDefaultMaybe = settings
  ? { default: settings.REACT_APP_SHARETRIBE_SDK_CLIENT_ID }
  : {};

// and using it:
return [
  {
    type: 'input',
    name: 'REACT_APP_SHARETRIBE_SDK_CLIENT_ID',
    message: 'blaa',
    validate: () => true,
    ...clientIdDefaultMaybe,
  }
];
```

scripts/config.js
```diff
@@ -0,0 +1,401 @@
...
+};
+
+/**
+ * Maps trough array of lines to create contet that is written to .env file. This will overwrite the previous content.
+ *
+ * @param {array} data arry of lines to be written into .env file
+ */
+const updateEnvFile = data => {
+  let content = '';
+  data.map(line => {
```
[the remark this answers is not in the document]
Gnito: Yeah, in this case `fs.writeFileSync('./.env', data.join(''))` might be enough. 

Most of the time `let someVariable` +  `map` statements can be done with plain `map` or with `reduce` function.

scripts/config.js
```diff
@@ -0,0 +1,398 @@
...
+      when: function(answers) {
+        return answers.showAdvancedSettings;
+      },
+    },
+    {
+      type: 'confirm',
+      name: 'REACT_APP_AVAILABILITY_ENABLED',
+      message: `Do you want to enable availability calendar?
+${chalk.dim(
+        'This setting enables the Availability Calendar for listings. The default value for this setting is true.'
```
Gnito: @OtterleyW sorry, I have completely forgotten that this environment variable exists. So, there's a bug - not a critical one, but definitely this wording doesn't make sense atm. My bad.

Before merging this PR, could you add a bug-fix to availability calendar code - it could be a part of this PR or added as a separate one.

What should be done:

**config.js** (expose new config):
```js
// Show availability calendar on EditListingWizard
const showAvailabilityCalendar = process.env.REACT_APP_AVAILABILITY_ENABLED === 'true';
```
**+** exposing it at the end of the file

**EditListingWizard.js** (conditionally show tab):
```js
```
```js
const availabilityMaybe = config.showAvailabilityCalendar ? [AVAILABILITY] : [];
export const TABS = [DESCRIPTION, FEATURES, POLICY, LOCATION, PRICING, ...availabilityMaybe, PHOTOS];
```

**ManageListingCard.js** (conditionally show "manage listing" link):
```js
{config.showAvailabilityCalendar
  ? (
    <React.Fragment>
      <span className={css.manageLinksSeparator}>{' • '}</span>

      <NamedLink
        className={css.manageLink}
        name="EditListingPage"
        params={{ id, slug, type: editListingLinkType, tab: 'availability' }}
      >
        <FormattedMessage id="ManageListingCard.manageAvailability" />
      </NamedLink>
    </React.Fragment>
    )
  : null}
```

scripts/config.js
```diff
@@ -0,0 +1,418 @@
...
+    } else {
+      return `${line}\n`;
+    }
+  });
+
+  return data;
+};
+
+/**
+ * Maps trough array of lines to create contet that is written to .env file. This will overwrite the previous content.
```
lyyder: This could be updated to match the current implementation.

PR #999, opened 2019-01-15
Update sentry

OtterleyW:
Update Sentry SDKs for browser and Node.js to version 4.5.1. With the new SDKs only one
  DSN needs to be configured so update also environment variables and documentation related to
  Sentry.


src/util/log.js
```diff
@@ -53,9 +57,17 @@ export const clearUserId = () => {
...
+    const extra = { ...data, apiErrorData: responseApiErrorInfo(e), code };
+
+    Sentry.withScope(scope => {
+      Object.keys(extra).map(tag => {
+        return scope.setTag(tag, extra[tag]);
+      });
+      Sentry.captureException(e);
+    });
+
+    //Sentry.captureException(e, { tags: { code }, extra });
```
Gnito: Remove this line.

src/util/log.js
```diff
@@ -53,9 +57,17 @@ export const clearUserId = () => {
...
-    Raven.captureException(e, { tags: { code }, extra });
+  if (config.sentryDsn) {
+    const extra = { ...data, apiErrorData: responseApiErrorInfo(e), code };
+
+    Sentry.withScope(scope => {
+      Object.keys(extra).map(tag => {
+        return scope.setTag(tag, extra[tag]);
+      });
+      Sentry.captureException(e);
+    });
```
Gnito: I think this should use `setExtra` instead of just `setTag`:
```js
const extra = { ...data, apiErrorData: responseApiErrorInfo(e) };

Sentry.withScope(scope => {
  scope.setTag("code", code);
  Object.keys(extra).forEach(key => {
    scope.setExtra(key, extra[key]);
  });
  Sentry.captureException(e);
});
```
- Usage in [React example](https://docs.sentry.io/platforms/javascript/react/)
- [set_extra in unified API](https://docs.sentry.io/development/sdk-dev/unified-api/#scope)
- https://docs.sentry.io/enriching-error-data/scopes/?platform=browsernpm#local-scopes

Gnito: Change this part also in server/log.js!

PR #1002, opened 2019-01-17
Enable limiting location autocomplete to a certain country

OtterleyW:
Limit location autocomplete by adding an optional country parameter to geocoding call in both Mapbox and Google Maps integrations. Variable used to enable this is `config.maps.search.limit` and it can be found from `config.js` file. The same variable works for Mapbox and Google Maps integrations.

Also updated Mapbox SDK to version 0.5.0.


src/config.js
```diff
@@ -129,6 +129,11 @@ const maps = {
     // `default-location-searches.js` file.
     defaults:
       process.env.REACT_APP_DEFAULT_SEARCHES_ENABLED === 'true' ? defaultLocationSearches : [],
+
+    // Limit location autocomplete to a one or more countries
+    // using ISO 3166 alpha 2 country codes separated by commas.
+    // If you don't want to limit the autocomplete you can comment out this value.
+    limit: ['AU'],
```
Gnito: Maybe changing the name to `countryLimit`

src/util/googleMaps.js
```diff
@@ -66,27 +66,32 @@ const predictionSuccessful = status => {
...
  * with getPlaceDetails call
+ * @param {Object} restriction - defines the component restrictions that can be used with
+ * the autocomplete service. Used to restrict search to specific country (or countries).
  *
  * @return {Promise<{ search, predictions[] }>} - Promise of an object
  * with the original search query and an array of
  * `google.maps.places.AutocompletePrediction` objects
  */
-export const getPlacePredictions = (search, sessionToken) =>
+export const getPlacePredictions = (search, sessionToken, restriction) =>
```
Gnito: It is probably better to rename this to `searchConfigurations` (or something like that) since there are more configurations that could be passed in than just `componentRestrictions`

src/components/LocationAutocompleteInput/GeocoderGoogleMaps.js
```diff
@@ -45,7 +45,11 @@ class GeocoderGoogleMaps {
    * only relevant for the `getPlaceDetails` function below.
    */
   getPlacePredictions(search) {
-    return getPlacePredictions(search, this.getSessionToken()).then(results => {
+    const limitCountriesMaybe = config.maps.search.limit
+      ? { country: config.maps.search.limit }
+      : {};
+    const restriction = { componentRestrictions: { ...limitCountriesMaybe } };
```
Gnito: Let's not send a key `componentRestrictions` if there is not `countryLimit` set.
```js
 const limitCountriesMaybe = config.maps.search.limit
      ? { 
           componentRestrictions: {
             country: config.maps.search.limit,
           },
         }
      : {};
```
Then `limitCountriesMaybe` could be spread directly to the call:
```js
return getPlacePredictions(search, this.getSessionToken(), limitCountriesMaybe).then(results => {
```

src/config.js
Gnito: And don't leave this `countryLimit` active!

PR #1005, opened 2019-01-18
Default to English translation if translation key is missing

OtterleyW:
If the locale in `config.js` is different than `en` we compare the selected translation to default `en.json` file.  Translations are compared in the same way that it's done in Translations CLI. 

After this update, new translation keys will not be added to other translation files with English default texts. We keep providing translations in our supported languages but they might not be up to date all the time. This means if you want to update your translations beforehand or use your own translations file, you can use [translation CLI]((https://github.com/sharetribe/flex-template-web/blob/master/docs/translations.md#managing-translations)) to check if there are translations missing.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -9,6 +9,7 @@ import 'react-dates/initialize';
 import Helmet from 'react-helmet';
 import { BrowserRouter, StaticRouter } from 'react-router-dom';
 import { Provider } from 'react-redux';
+import difference from 'lodash/difference';
 import mapValues from 'lodash/mapValues';
 import moment from 'moment';
 import { IntlProvider, addLocaleData } from 'react-intl';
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -9,7 +9,7 @@ import 'react-dates/initialize';
 import Helmet from 'react-helmet';
 import { BrowserRouter, StaticRouter } from 'react-router-dom';
 import { Provider } from 'react-redux';
-import mapValues from 'lodash/mapValues';
+import { difference, mapValues } from 'lodash';
```
Gnito: Don't export anything directly from Lodash. 
It is quite a big library and since we use a very small portion of functions it provides, we save almost 15 KB of gzipped javascript if we don't import it directly.

src/app.js
```diff
@@ -18,15 +18,39 @@ import Routes from './Routes';
...
+if (config.locale !== 'en') {
+  const sourceKeys = Object.keys(defaultMessages);
+  const targetKeys = Object.keys(messagesInLocale);
+  const diff = difference(sourceKeys, targetKeys);
+
+  const addedKeys = diff.reduce(function(toAdd, key) {
+    const object = {};
+    object[key] = defaultMessages[key];
+    return Object.assign({}, toAdd, object);
+  }, {});
```
Gnito: Use arrow function instead of anonymous `function` declaration and I think object spread is easier to understand than `Object.assign`.
With those changes this can be turned to oneliner:
```js
const addedKeys = diff.reduce((toAdd, key) => ({ ...toAdd, [key]: defaultMessages[key] }), {});
```

or if you prefer explicit returns:
```js
const addedKeys = diff.reduce((toAdd, key) => {
  return { ...toAdd, [key]: defaultMessages[key] };
}, {});
```

Although, this is probably the easiest to understand:
```js
  const addDefaultTranslation = (translations, missingKey) => ({ ...translations, [missingKey]: defaultMessages[missingKey] });
  const addedKeys = diff.reduce(addDefaultTranslation, messagesInLocale);
```

Gnito: Actually, I'd suggest this:
```js
const addMissingTranslations = (sourceLangTranslations, targetLangTranslations) => {
  const sourceKeys = Object.keys(sourceLangTranslations);
  const targetKeys = Object.keys(targetLangTranslations);
  const missingKeys = difference(sourceKeys, targetKeys);

  const addDefaultTranslation = (translations, missingKey) => 
    ({ ...translations, [missingKey]: sourceLangTranslations[missingKey] });

  return missingKeys.reduce(addDefaultTranslation, targetLangTranslations);
};

const isDefaultLanguageInUse = config.locale === 'en';
const messages = isDefaultLanguageInUse
  ? defaultMessages
  : addMissingTranslations(defaultMessages, messagesInLocale);
```

Gnito: (btw. I didn't test that ^)

src/app.js
```diff
@@ -18,15 +18,39 @@ import Routes from './Routes';
...
+import messagesInLocale from './translations/en.json';
+
 // If you are using a non-english locale with moment library,
 // you should also import time specific formatting rules for that locale
 // e.g. for French: import 'moment/locale/fr';
 
+// Flex template application uses English translations as default.
+// If translation key is missing from e.g. fr.json corresponding key
+// will be added to messages from en.json to prevent missing translation key errors.
+import defaultMessages from './translations/en.json';
```
Gnito: Now that I read these comments, I think that these lines should be reordered a bit:

```js
// Flex template application uses English translations as default.
import defaultMessages from './translations/en.json';

// If you want to change the language, change the imports to match the wanted locale:
//   1) Change the language in the config.js file!
//   2) Import correct locale rules for React Intl library
//   3) Import locale rules for Moment library
//   4) Use the `messagesInLocale` import to add the translation file.

// Step 2:
// Import locale rules for React Intl library 
import localeData from 'react-intl/locale-data/en';

// Step 3:
// If you are using a non-english locale with moment library,
// you should also import time specific formatting rules for that locale
// e.g. for French: import 'moment/locale/fr';

// Step 4:
// If you are using a non-english locale, point `messagesInLocale` to correct .json file
import messagesInLocale from './translations/en.json';

// If translation key is missing from `messagesInLocale` (e.g. fr.json), 
// corresponding key will be added to messages from `defaultMessages` (en.json) 
// to prevent missing translation key errors.
```

Gnito: Also docs should reflect this step-by-step model ^

PR #1009, opened 2019-01-24
Use the same FSM description format as Xstate library

Gnito:


src/util/transaction.js
```diff
@@ -52,81 +61,207 @@ export const TRANSITIONS = [
...
+const STATE_REVIEWED = 'reviewed';
+const STATE_REVIEWED_BY_CUSTOMER = 'reviewed-by-customer';
+const STATE_REVIEWED_BY_PROVIDER = 'reviewed-by-provider';
 
-export const transitionsToRequested = [TRANSITION_REQUEST, TRANSITION_REQUEST_AFTER_ENQUIRY];
-export const txIsRequested = tx => transitionsToRequested.includes(txLastTransition(tx));
+/**
+ * Description of transaction process
+ *
+ * You should keep this in sync with transaction process defined Flex API
```
OtterleyW: "defined in Marketplace API"

Flex API should be called Marketplace API like in [Flex concepts](https://www.sharetribe.com/docs/background/concepts/)

src/util/transaction.js
```diff
@@ -52,81 +61,207 @@ export const TRANSITIONS = [
...
+        [TRANSITION_REVIEW_2_BY_CUSTOMER]: STATE_REVIEWED,
+        [TRANSITION_EXPIRE_CUSTOMER_REVIEW_PERIOD]: STATE_REVIEWED,
+      },
+    },
+    [STATE_REVIEWED]: { type: 'final' },
+  },
+};
+
+// Note: currently we assume that state config doesn't contain nested states.
+const statesFromConfig = config => config.states || {};
```
OtterleyW: I first thought `config` meant config.js file so could we use `stateConfig` here too?

src/util/transaction.js
```diff
@@ -52,81 +61,215 @@ export const TRANSITIONS = [
...
+/**
+ * Description of transaction process
+ *
+ * You should keep this in sync with transaction process defined in Marketplace API
+ *
+ * Note: we don't use yet any state machine library,
+ * but this description format is following Xstate (FSM library)
+ */
+const stateDescription = {
+  id: 'preauth-nightly-booking',
```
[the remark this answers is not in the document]
Gnito: Changed to:
```js
  // id is defined only to support Xstate format.
  // However if you have multiple transaction processes defined,
  // it is best to keep them in sync with transaction process aliases.
  id: 'preauth-with-nightly-booking/release-1',
```

PR #1010, opened 2019-01-24
Improve activity feed

OtterleyW:


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -87,46 +131,17 @@ Review.propTypes = {
...
-  return [
-    TRANSITION_REVIEW_1_BY_CUSTOMER,
-    TRANSITION_REVIEW_1_BY_PROVIDER,
-    TRANSITION_REVIEW_2_BY_CUSTOMER,
-    TRANSITION_REVIEW_2_BY_PROVIDER,
-  ].includes(transition);
-};
-
-const hasUserLeftAReviewFirst = (userRole, lastTransition) => {
+const hasUserLeftAReviewFirst = (userRole, transaction, lastTransition) => {
```
Gnito: I think this should be separate utility function in util/transaction.js

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -87,46 +131,17 @@ Review.propTypes = {
...
+    (isProvider && txHasFirstReviewBy(transaction, isProvider)) ||
+    (!isProvider && txHasFirstReviewBy(transaction, !isProvider)) ||
     areReviewsCompleted(lastTransition)
   );
 };
 
 const resolveTransitionMessage = (
+  transaction,
   transition,
   lastTransition,
```
Gnito: Remove this one. It is not needed since transaction is passed in.

Gnito: Also remove line 156 (`deliveredState`) and, in line 193, change `deliveredState` to `txIsCompleted(transaction)`
(Although, in my subbranch it will be renamed as `txIsDelivered`.)

src/components/ActivityFeed/ActivityFeed.js
Gnito: I think this would be more readable with variables:

```js
const hasUserLeftAReviewFirst = (userRole, transaction) => {
  const isProvider = txRoleIsProvider(userRole);
  return isProvider
    ? txHasFirstReviewBy(transaction, isProvider));
    : txHasFirstReviewBy(transaction, !isProvider));
};
```
then usage should be something like:

```js
    case TRANSITION_COMPLETE: {
      // Show the leave a review link if the state is delivered or
      // if current user is not the first to leave a review
      const reviewPeriodJustStarted = txIsDelivered(transaction);
      const reviewPeriodIsOver = txIsReviewed(transaction);
      const userHasLeftAReview = hasUserLeftAReviewFirst(ownRole, transaction);
      const reviewLink =
        reviewPeriodJustStarted || !(reviewPeriodIsOver || userHasLeftAReview) ? (
          <InlineTextButton onClick={onOpenReviewModal}>
            <FormattedMessage id="ActivityFeed.leaveAReview" values={{ displayName }} />
          </InlineTextButton>
        ) : null;

      return <FormattedMessage id="ActivityFeed.transitionComplete" values={{ reviewLink }} />;
    }
```
```js
        // show the leave a review link if current user is not the first
        // one to leave a review
        const reviewPeriodIsOver = txIsReviewed(transaction);
        const userHasLeftAReview = hasUserLeftAReviewFirst(ownRole, transaction);
        const reviewLink = !(reviewPeriodIsOver || userHasLeftAReview) ? (
```

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -175,7 +190,7 @@ const resolveTransitionMessage = (
       // Show the leave a review link if the state is delivered or
       // if current user is not the first to leave a review
       const reviewLink =
-        deliveredState || !hasUserLeftAReviewFirst(ownRole, lastTransition) ? (
+        deliveredState || !hasUserLeftAReviewFirst(ownRole, transaction, lastTransition) ? (
```
Gnito: There's another call for `hasUserLeftAReviewFirst` on line 207.

PR #1011, opened 2019-01-25
en translation fix for ReviewForm.reviewRatingRequired

Schwerbelastung:


lyyder:
It seems that this message is never visible, which is probably why it's been wrong all along. It's used in the `ReviewForm` component and I don't think the submit button for the form is enabled before the rating value is provided. And there is now way to focus the review start input and then leave it empty, which would render this message. This is because it's impossible to "empty" the rating input and it can't be navigated into using TAB.

Gnito:
@lyyder, yeah. It is an accessibility problem. There should be `tabindex` in the `<label>` elements + some keyboard listener attached to them too.

What comes to error message: our template app doesn't need it but it's good to have error handling ready in every Field* input.

Schwerbelastung:
And it's one less mistake that any outside can find in our code :)

PR #1019, opened 2019-02-01
Page content width should be enforced to 1056px

OtterleyW:
If the page is as wide as footer use same way to count the page width.

This will affect e.g.`LandingPage`, `AboutPage`, `ListingPage`and `TransactionPanel` and the side navigation e.g. `LayoutWrapperSideNav`, `InboxPage` and `EditListingWizard`.

### DIFF
--- a/src/components/EditListingWizard/EditListingWizard.css
+++ b/src/components/EditListingWizard/EditListingWizard.css
@@ -38,12 +38,16 @@
   overflow-x: scroll;
 
   @media (--viewportLarge) {
-    padding: 128px 0 82px calc((100% - 1056px) / 2 + 74px);
+    padding: 128px 0 82px 36px;
     flex-direction: column;
     overflow-x: auto;
     background-color: var(--matterColorBright);
     box-shadow: none;
     border-top: none;
+  }
+
+  @media (--viewportLargeWithPaddings) {
+    padding: 128px 0 82px calc((100% - 1056px) / 2);
   }
 }
 
@@ -80,9 +84,13 @@
   flex-grow: 1;
 
   @media (--viewportLarge) {
-    padding: 88px calc((100% - 1056px) / 2 + 36px) 82px 82px;
+    padding: 88px 36px 82px 59px;
     border-left: 1px solid var(--matterColorNegative);
     background-color: var(--matterColorLight);
+  }
+
+  @media (--viewportLargeWithPaddings) {
+    padding: 88px calc((100% - 1056px) / 2) 82px 59px;
   }
 }
 
--- a/src/components/EditListingWizard/EditListingWizard.css
+++ b/src/components/EditListingWizard/EditListingWizard.css
@@ -38,17 +38,12 @@
   overflow-x: scroll;
 
   @media (--viewportLarge) {
-    padding: 126px 0 82px 15vw;
+    padding: 128px 0 82px calc((100% - 1056px) / 2 + 74px);
     flex-direction: column;
     overflow-x: auto;
-
     background-color: var(--matterColorBright);
     box-shadow: none;
     border-top: none;
-  }
-
-  @media (--viewportXLarge) {
-    padding-left: 25vw;
   }
 }
 
@@ -85,13 +80,9 @@
   flex-grow: 1;
 
   @media (--viewportLarge) {
-    padding: 82px 15vw 82px 82px;
+    padding: 88px calc((100% - 1056px) / 2 + 36px) 82px 82px;
     border-left: 1px solid var(--matterColorNegative);
     background-color: var(--matterColorLight);
-  }
-
-  @media (--viewportXLarge) {
-    padding-right: 25vw;
   }
 }
 
--- a/src/components/Footer/Footer.css
+++ b/src/components/Footer/Footer.css
@@ -10,7 +10,7 @@
   background-color: var(--matterColorBright);
 
   @media (--viewportMedium) {
-    padding: 0 30px 33px 30px;
+    padding: 0 24px 33px 24px;
   }
 
   @media (--viewportLarge) {
@@ -41,12 +41,17 @@
   width: 100%;
 
   @media (--viewportMedium) {
-    max-width: 1056px;
     margin: 0 auto;
   }
 
   @media (--viewportLarge) {
-    padding: 0 24px;
+    max-width: 1128px;
+    padding: 0 36px;
+  }
+
+  @media (--viewportXLarge) {
+    max-width: 1056px;
+    padding: 0;
   }
 }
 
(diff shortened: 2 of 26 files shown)


src/components/LayoutWrapperSideNav/LayoutWrapperSideNav.css
```diff
@@ -14,17 +14,13 @@
   box-shadow: var(--boxShadow);
 
   @media (--viewportLarge) {
-    padding: 112px 0 82px 15vw;
+    padding: 112px 0 82px calc((100% - 1056px) / 2 + 74px);
```
Gnito: This is part looks good 👍 

However, I noticed that subelements have still strange values. 

I'd propose that we also update `.tabs` in this file and `.link` in TabNav.

```css
@media (min-width: 1024px)
.LayoutWrapperSideNav__tabs__EKdgG {
  min-height: auto;
  flex-direction: column;
  margin-top: 24px; /* 28px is not divisible by baseline 8px */
  padding-top: 0;
}
```

Currently, in TabNav, we have correct height 56px (56/8=7). But it is constructed poorly:
```css
@media (min-width: 1024px)
.TabNav__link__1PgdT {
  border-bottom-width: 0px;
  margin-bottom: 14px;
  width: 200px;
  padding-top: 0;
  margin-top: 0;
  /* and padding-bottom: 10px; is coming from previous breakpoint */
}

```
It should be something like:
```css
@media (min-width: 1024px)
.TabNav__link__1PgdT {
  width: 200px;
  border-bottom-width: 0px;
  margin-top: 0;
  margin-bottom: 16px; /* 16/8=2 */

  /* Line-height is 32px and adding padding of 6px + 2px gives a total height of 40px which divisible by 8 */
  padding-top: 6px; /* Drop the text to adjust it to correct baseline */
  padding-bottom: 2px; 
}
```

Wurthermore, we also have problems with LayoutSideNavigation:
```css
@media (min-width: 1024px)
.LayoutSideNavigation__layoutWrapperMain__3jQ70 {
  padding: 116px 15vw 82px 82px; /* 116 / 8 = 14.5 => I think this should also be 112px;*/
}
```

That would mean that we should change `.heading` on TermsOfServicePage:
```css
@media (min-width: 768px)
.TermsOfServicePage__heading__2ni4I {
  margin: 8px 0 24px 0; /* currently, this is "margin: 2px 0 23px 0;" */
}
```
and in several other pages need similar adjustment since they use `LayoutSideNavigation`: 
- PrivacyPolicy
- ContactDetailsPage
- InboxPage
- PasswordChangePage
- PayoutPreferencesPage
- ProfilePage

src/components/Footer/Footer.css
```diff
@@ -41,13 +41,18 @@
   width: 100%;
 
   @media (--viewportMedium) {
-    max-width: 1056px;
     margin: 0 auto;
   }
 
   @media (--viewportLarge) {
+    max-width: 1104px;
```
Gnito: There's still mismatch (at least) on LandingPage with paddings. Topbar and SectionHero use 36px padding but content and Footer use 24px. That is pretty clear on one quite important breakpoint 1024px, which is common width on tablets.

So, I'd push this change a bit further and recommend paddings to be 36px starting from that break
screenshot 2019-02-04 at 14 36 47
point.

src/components/EditListingWizard/EditListingWizard.css
```diff
@@ -85,14 +80,10 @@
   flex-grow: 1;
 
   @media (--viewportLarge) {
-    padding: 82px 15vw 82px 82px;
+    padding: 88px calc((100% - 1056px) / 2 + 36px) 82px 82px;
```
Gnito: Unfortunately, there are still problems with this.

`calc((100% - 1056px) / 2 + 36px)`
means that the right column will always be 36px narrower than Footer.
That's because `calc((100% - 1056px) / 2)` is already the filling the distance between browser window's right edge and 1056px wide content area.

However, we didn't think through what happens at the breakpoint. If the `window.innerWidth` is 1024px (aka `--viewportLarge`), then this calculation is actually `calc((1024px - 1056px) / 2)` which means `-16px` and negative padding is a faulty value for browsers and they (at least Chrome) discard it altogether.

This is rather annoying problem:
I couldn't think of any other solution than to add a new breakpoint for this component so that layout between **1024px** and **1128px** are covered differently than larger layouts. 
```css
@media (--viewportLarge) {
  padding: 88px 36px 82px 82px;
```
and 
```css
@media (min-width: 1128px) {
  padding: 88px calc((100% - 1056px) / 2) 82px 82px;
}
```

Similar breakpoint should be for the sidenav
`padding: 112px 0 82px 36px;` vs `padding: 112px 0 82px calc((100% - 1056px) / 2 + 74px);`

----

I made a quick search if we have used extra breakpoints anywhere and it seems that there are only 2 components that has wild breakpoints:
- **TopbarDesktop.css** uses `@media (min-width: 1024px) {` 
  - we could just change it to `@media (--viewportLarge) {`
- **ManageAvailabilityCalendar.css** uses `@media (min-width: 550px) {`
  - My recent addition since content in React-Dates calendar didn't work well with small table-cells.

Therefore, my suggestion is actually that we:
1) fix Topbar.css and 
2) add two new custom breakpoints:
  - `@custom-media --viewportSmall (min-width: 550px);`
  - `@custom-media --viewportLargeWithPaddings (min-width: 1128px);`

Gnito:
One more thing in the comments.

After those changes, I suggest that we let this branch to wait for design review (which should happen next week...)

Remember to test with other browsers too.

src/components/LayoutSideNavigation/LayoutSideNavigation.css
```diff
@@ -35,14 +35,14 @@
...
     background-color: var(--matterColorLight);
     border-left-width: 1px;
     border-left-style: solid;
     border-left-color: var(--matterColorNegative);
   }
 
-  @media (--viewportXLarge) {
-    padding-right: 25vw;
+  @media (--viewportLargeWithPaddings) {
+    padding: 112px calc((100% - 1056px) / 2 + 36px) 82px 82px;
```
Gnito: If you keep the `+ 36px` after viewportLargeWithPaddings, it will shrink the content smaller than 1056px:
screenshot 2019-02-06 at 12 10 05

Gnito: This applies to the right side. On the left side we want to achieve this effect since the content is fixed 200px wide and the reserved space is bigger.

src/containers/LandingPage/LandingPage.css
```diff
@@ -85,13 +85,19 @@
   margin: var(--LandingPage_sectionMarginTop) 24px 51px 24px;
 
   @media (--viewportMedium) {
-    margin: var(--LandingPage_sectionMarginTopMedium) 7.5vw 60px 7.5vw;
+    max-width: 100%;
+    margin: var(--LandingPage_sectionMarginTopMedium) 30px 60px 30px;
```
Gnito: I was a bit worried about what's going on here (since Topbar uses 24px horizontal paddings), but I guess, we need to ask from Janne: `(Janne Koivistoinen 2017-12-22 15:33:19 +0200 13)     padding: 0 30px 33px 30px;`

PR #1020, opened 2019-02-04
Add audit script to check dependency vulnerabilities

Gnito:
The output of `yarn run audit` when it finds new vulnerabilities is:

screenshot 2019-02-04 at 21 42 27

> Note: bfj (big-friendly-JSON) dependency is part of CRA 2.x too.


scripts/audit.js
```diff
@@ -0,0 +1,90 @@
...
+
+// get an advisory or empty object
+const getAdvisory = o => (!o ? {} : !o.data ? {} : !o.data.advisory ? {} : o.data.advisory);
+// get a resolution or empty object
+const getResolution = o => (!o ? {} : !o.data ? {} : !o.data.resolution ? {} : o.data.resolution);
+
+// Read exceptions
+let exceptions = [];
+bfj
+  .parse(fs.createReadStream('.nsprc'), {})
```
kpuputti: You can also just import a JSON file.

.nsprc
```diff
@@ -1,10 +1,8 @@
...
-    "https://nodesecurity.io/advisories/532",
-    "https://nodesecurity.io/advisories/157",
-    "https://nodesecurity.io/advisories/577",
-    "https://nodesecurity.io/advisories/654",
-    "https://nodesecurity.io/advisories/664",
-    "https://nodesecurity.io/advisories/678"
+    "https://npmjs.com/advisories/532",
+    "https://npmjs.com/advisories/577",
+    "https://npmjs.com/advisories/722",
+    "https://npmjs.com/advisories/725"
```
kpuputti: Should this file be renamed if it's a custom format now? The data could even be inlined in the audit script with proper comments. This used to have comments on why the exceptions were there, and that was very helpful (e.g. dev server vulnerability is ok).

scripts/audit.js
Gnito: I didn't want to save any file anywhere. It's a huge file - that's pretty much unreadable for humans due to the size of printed JSON.

Gnito: This was different import :)

I just decided that it's better to import both the same way. Plus I thought that it might be a good discussion point if the exception list should actually be `.auditrc` file with JavaScript syntax since that would allow comments (unlike JSON).

kpuputti: Are you talking about a different file?

.nsprc
Gnito: Just one clarification: it _never_ had comments since the original format `.nsprc` was JSON too.

I want to keep this separated from the actual audit script since the implementation is useless detail for most customizers, but updating the file would cause conflicts.

How about renaming the file to `.auditrc` (audit runtime configuration). I was also thinking of using JavaScript instead of JSON, which would allow comments.

scripts/audit.js
kpuputti: At least the `.nsprc` name is misleading as NSP isn't visible being used at all. Comments in JSON work fine in Node, but it's a subjective decision if you don't want to do that.

Anyways, I trust your decision here, no need to go back and forth any further.

.nsprc
kpuputti: The original version did have comments:

https://github.com/sharetribe/flex-template-web/commit/eb9c909939ec0b06c6dd044e832a57dfcd4cebe3#diff-03cb303c8d3085901cf5a9214888ad1d

They just changed the parser at some point.

scripts/audit.js
Gnito: Didn't know about node allowing comments :o. I was actually annoyed by that. (However, I'm not sure if bfj allows comments.

.nsprc
Gnito: :)

Gnito: Github is quite bad as a chat platform :D

Anyway, good feedback. I was hoping that you would comment about this file. I was trying to decide between following previous format or using JS (due to comments).

Anyway, after this (confusingly async) chat, I'll rename this file: `.auditrc` and use JavaScript inside of it.

PR #1022, opened 2019-02-07
Show deleted user

OtterleyW:
A new component `UserDisplayName` is added for showing user display name and also handling the cases where a user is banned or deleted. When the user name must be a string instead of a component (e.g. in `Avatar` and in `ListingPage`) you can use a new function `userDisplayNameAsString`. Together these will replace the old `userDisplayName` function which is now deprecated. There is quite a lot of file changes caused by updating test files.

This PR contains also small bug fixes:
- prop type `user` now accepts also banned or deleted user
- image size in mobile view when a listing doesn't have an image


src/components/Avatar/Avatar.js
```diff
@@ -34,14 +34,26 @@ export const AvatarComponent = props => {
   const avatarUser = userIsCurrentUser ? ensureCurrentUser(user) : ensureUser(user);
 
   const isBannedUser = avatarUser.attributes.banned;
+  const isDeletedUser = avatarUser.attributes.banned;
```
Gnito: Copy-paste bug: `const isDeletedUser = avatarUser.attributes.banned;`

src/util/data.js
```diff
@@ -238,18 +238,45 @@ export const ensureAvailabilityException = availabilityException => {
...
+  const hasProfile = hasAttributes && user.attributes.profile;
+
+  if (isBanned || isDeleted) {
+    return defaultUserDisplayName;
+  } else if (hasProfile) {
+    return user.attributes.profile.displayName;
+  } else {
+    return '';
+  }
+};
```
Gnito: I'd suggest that this is turned the other way around:
```js
if (hasProfile) {
  return user.attributes.profile.displayName;
} else {
  return defaultUserDisplayName || '';
}
```

Then there's no need to check if the user is banned or deleted or anything else.

src/util/data.js
```diff
@@ -238,18 +238,45 @@ export const ensureAvailabilityException = availabilityException => {
...
  * @param {propTypes.user} user
  * @param {String} bannedUserDisplayName
  *
  * @return {String} display name that can be rendered in the UI
  */
 export const userDisplayName = (user, bannedUserDisplayName) => {
+  console.warn(
+    `Function userDisplayName is deprecated! 
+User function userDisplayNameAsString or component UserDisplayName instead.`
+  );
```
Gnito: You could change this so that it calls internally `userDisplayNameAsString`

src/util/data.js
```diff
@@ -267,20 +294,22 @@ export const userDisplayName = (user, bannedUserDisplayName) => {
...
-export const userAbbreviatedName = (user, bannedUserAbbreviatedName) => {
+export const userAbbreviatedName = (user, defaultUserAbbreviatedName) => {
   const hasAttributes = user && user.attributes;
   const hasProfile = hasAttributes && user.attributes.profile;
+  const isBanned = hasAttributes && user.attributes.banned;
+  const isDeleted = hasAttributes && user.attributes.deleted;
 
-  if (hasAttributes && user.attributes.banned) {
-    return bannedUserAbbreviatedName;
+  if (isBanned || isDeleted) {
```
Gnito: I think this should be swapped too.
```js
if (hasAbbreviatedName) {
  return user.attributes.profile.abbreviatedName;
} else {
  return defaultUserAbbreviatedName || '';
}
```

src/util/types.js
```diff
@@ -109,18 +109,28 @@ propTypes.currentUser = shape({
   profileImage: propTypes.image,
 });
 
+const userAttributes = shape({
+  banned: bool.isRequired,
```
Gnito:
```js
  banned: propTypes.value(false).isRequired,
  deleted: propTypes.value(false).isRequired,
```

src/util/types.js
```diff
@@ -109,18 +109,28 @@ propTypes.currentUser = shape({
...
   type: propTypes.value('user').isRequired,
-  attributes: shape({
-    banned: bool.isRequired,
-    profile: shape({
-      displayName: string.isRequired,
-      abbreviatedName: string.isRequired,
-      bio: string,
-    }),
-  }),
+  attributes: oneOfType([userAttributes, deletedUserAttributes, bannedUserAttributes]).isRequired,
```
Gnito: 👍

src/components/UserDisplayName/UserDisplayName.js
```diff
@@ -0,0 +1,47 @@
...
+        id: 'UserDisplayName.deleted',
+      });
+
+  const bannedUserDisplayNameInUse = bannedUserDisplayName
+    ? bannedUserDisplayName
+    : intl.formatMessage({
+        id: 'UserDisplayName.banned',
+      });
+
+  return userIsDeleted ? (
```
Gnito: I think this would be more readable in a format:
```js
  const classes = classNames(rootClassName, className);
  return (
    <span className={classes}>
      {userHasProfile
        ? user.attributes.profile.displayName
        : userIsDeleted
        ? deletedUserDisplayNameInUse
        : userIsBanned
        ? bannedUserDisplayNameInUse
        : null}
    </span>
  );
```

src/components/TransactionPanel/TransactionPanel.css
```diff
@@ -86,6 +86,8 @@
 
 /* Passed-in props for DetailCardImage subcomponent */
 .imageWrapperMobile {
+  position: relative;
```
Gnito: 👍

src/components/TransactionPanel/TransactionPanel.js
```diff
@@ -236,11 +235,11 @@ export class TransactionPanelComponent extends Component {
...
       currentProvider,
       currentCustomer,
-      bannedUserDisplayName
+      intl
     );
 
-    const { publicData = {}, geolocation } = currentListing.attributes;
-    const location = publicData.location || {};
+    const { publicData, geolocation } = currentListing.attributes;
+    const location = publicData && publicData.location ? publicData.location : {};
```
Gnito: 👍

src/components/TransactionPanel/TransactionPanel.js
```diff
@@ -334,17 +336,19 @@ export class TransactionPanelComponent extends Component {
               onShowMoreMessages={() => onShowMoreMessages(currentTransaction.id)}
               totalMessagePages={totalMessagePages}
             />
+            {showSendMessageForm ? (
```
Gnito: This is great. 👍 

Maybe there could be an info text that states why you can't send messages atm.

PR #1035, opened 2019-02-26
Update creating Stripe accounts to match the updated Stripe API

OtterleyW:


src/stripe-config.js
```diff
@@ -7,6 +7,10 @@
 // To make Stripe connection work, you also need to set Stripe's private key in the Flex Console.
 export const stripePublishableKey = process.env.REACT_APP_STRIPE_PUBLISHABLE_KEY;
 
+// You can check your Stripe API version from Stripe Dashboard -> Developers.
+// If you are using the API version that is earlier than '2019-01-19' change this value to 'false'
+export const useNewStripeAPIversion = true;
```
Gnito: I like that this is a true/false flag since we don't support a wide variety of Stripe versions. However, I think we should tie the flag to this specific version or change. 

So, maybe something like 
```js
const useStripeAPIVersion20190219 = true;
``` 
or 
```js
const useOutdatedLegalEntityWithStripe = false;
```

In addition, change the API version to correct one ('2019-02-19') and we could also add a link to 
https://stripe.com/docs/upgrades#api-changelog

Furthermore, we should also mention that we don't yet support company accounts with '2019-02-19' and we should hide that option from PayoutSettingsForm (but let's discuss about that with Juho too).

PR #1042, opened 2019-03-06
Show error if Stripe publishable key is not configured

OtterleyW:
Show error in `PayoutDetailsForm`  and in `StripePaymentForm` if Stripe publishable key is not configured. This way the app will not crash e.g. if someone tries to publish a draft listing and the Stripe key is not configured yet. 

localhost_3000_l_test-sauna_5c78ed4e-d56c-46e7-8541-9504cc990650_draft_photos

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -753,6 +753,7 @@
   "StripePaymentForm.messageLabel": "Say hello to your host {messageOptionalText}",
   "StripePaymentForm.messageOptionalText": "• optional",
   "StripePaymentForm.messagePlaceholder": "Hello {name}! I'm looking forward to…",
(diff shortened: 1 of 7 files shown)


src/translations/en.json
```diff
@@ -552,6 +552,7 @@
   "PayoutDetailsForm.lastNameLabel": "Last name",
   "PayoutDetailsForm.lastNamePlaceholder": "Doe",
   "PayoutDetailsForm.lastNameRequired": "This field is required",
+  "PayoutDetailsForm.missingStripeKey": "You need to configure the Stripe publishable key to continue.",
```
Gnito: Providers see this message, so I think this should be changed to something like:
"Marketplace operator has not configured Stripe publishable key to this marketplace. Unfortunately, you can't publish your listing yet."

PR #1048, opened 2019-03-20
Fix empty slug

Gnito:
A listing title that contained only stripped-off characters caused bugs in slug / pathName generation.

Longer version:
If a provider creates a listing with a title that only contains characters that we strip-off when creating a slug for URLs, It will cause an error and a blank page (or 500 if server rendered). (E.g. title like "--")
This can happen in search page too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2019-XX-XX
 
+- [fix] A listing title that contained only stripped-off characters caused bugs in slug / pathName
+  generation. [#1048](https://github.com/sharetribe/flex-template-web/pull/1048)
 - [change] Removed Node-engine setup from package.json. Fixed version was causing problems for quite
   many in their first FTW installation. Note: when troubleshooting your Heroku installation, you
   might want to reintroduce engine setup.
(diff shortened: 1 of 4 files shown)


src/util/routes.js
```diff
@@ -20,8 +20,11 @@ const toPathByRouteName = (nameToFind, routes) => {
 /**
  * Shorthand for single path call. (```pathByRouteName('ListingPage', routes, { id: uuidX });```)
  */
-export const pathByRouteName = (nameToFind, routes, params = {}) =>
-  toPathByRouteName(nameToFind, routes)(params);
+export const pathByRouteName = (nameToFind, routes, params = {}) => {
+  const hasEmptySlug = params && Object.keys(params).find(k => k === 'slug') && params.slug === '';
```
kpuputti: No need to iterate the keys:

```suggestion
  const hasEmptySlug = params && params.hasOwnProperty('slug') && params.slug === '';
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2019-XX-XX
 
+- [fix] Title containing only stripped characters caused bugs in slug / pathName genetation.
```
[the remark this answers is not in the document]
Gnito: Changed to: "A listing title that contained only stripped-off characters caused bugs in slug / pathName generation."

src/util/routes.js
Gnito: Ah, true. I should stay in my bed.

PR #1050, opened 2019-03-25
Use displayStart and dislayEnd (Booking entity)

Gnito:


src/components/BookingBreakdown/LineItemBookingPeriod.js
```diff
@@ -40,9 +40,12 @@ const BookingPeriod = props => {
 const LineItemBookingPeriod = props => {
   const { transaction, booking, unitType } = props;
 
-  const { start: startDate, end: endDateRaw } = booking.attributes;
-  const localStartDate = dateFromAPIToLocalNoon(startDate);
-  const localEndDateRaw = dateFromAPIToLocalNoon(endDateRaw);
+  // Attributes: displayStart and displayEnd can be used to differentiate shown time range
+  // from actual start and end. It can help in situations where there are preparation time needed
```
bladealslayer: Perhaps s/actual start and end/actual start and end times used for availability reservation/?

src/components/BookingBreakdown/LineItemBookingPeriod.js
```diff
@@ -40,9 +40,12 @@ const BookingPeriod = props => {
...
+  // between bookings. Read more: https://www.sharetribe.com/api-reference/#bookings
+  const { start, end, displayStart, displayEnd } = booking.attributes;
+  const localStartDate = dateFromAPIToLocalNoon(displayStart || start);
```
bladealslayer: You'll always get back `displayStart` and `displayEnd` so the `||` isn't necessary here, I think.

src/containers/InboxPage/InboxPage.js
```diff
@@ -136,9 +136,12 @@ export const txState = (intl, tx, type) => {
...
 const bookingData = (unitType, tx, isOrder, intl) => {
-  const { start, end } = tx.booking.attributes;
-  const startDate = dateFromAPIToLocalNoon(start);
-  const endDateRaw = dateFromAPIToLocalNoon(end);
+  // Attributes: displayStart and displayEnd can be used to differentiate shown time range
+  // from actual start and end. It can help in situations where there are preparation time needed
+  // between bookings. Read more: https://www.sharetribe.com/api-reference/#bookings
+  const { start, end, displayStart, displayEnd } = tx.booking.attributes;
+  const startDate = dateFromAPIToLocalNoon(displayStart || start);
+  const endDateRaw = dateFromAPIToLocalNoon(displayEnd || end);
```
bladealslayer: Same comments as above.

src/components/BookingBreakdown/LineItemBookingPeriod.js
bladealslayer: At some point, it might make sense to change FTW to start setting display times when initiating transactions. That way the dateFromAPIToLocalNoon would become unnecessary here.

Gnito: @bladealslayer Yes, definitely. However, that means that we should include some kind of timezone info to extended data to make the behavior more understandable. So, it's probably better to do that when we start thinking about time-based bookings.

I added `||` since there's beta-flag in this feature. So, this shouldn't break even if API decides to change something later on.

PR #1058, opened 2019-03-29
react subdirectory support

lastlink:


Gnito:
Hi @lastlink, is this intentional pull request or an accidental one?


This seems like a useful feature, so I'd be happy to get this forward. 
However, there are a couple of things that should be addressed.

**1) `.env-template`**
 - `PUBLIC_URL`: why is this needed? It creates errors on server-side rendering.
 - We already have `REACT_APP_CANONICAL_ROOT_URL` root URL setting, which should be used for setting domain name ('https://yourwhitelabeldomain.com' or 'http://localhost:3000'), so that resources are pointing to a correct URI. (E.g. `<meta data-react-helmet="true" property="og:url" content="http://localhost:4000/subdir/l/5bda0a8c-c02e-43c7-8ef2-ae772a31727e">`)

So, my suggestion is that instead of `REACT_APP_BASENAME`, you just add a new comment next to `REACT_APP_CANONICAL_ROOT_URL` that 
```sh
# If you want to use a subdirectory, you need to include it to this variable (e.g.`http://localhost:4000/subdir`
# Note: "dev-server" script in package.json needs to be modified too.
```

**2) `src/config.js`**
- `basename` is not exported. However, it's not needed at all if `REACT_APP_CANONICAL_ROOT_URL` is used instead.

**3) `src/app.js`**
This doesn't add anything to server-side rendering -> and that causes lots of errors.

I'd suggest that you add a function to extract path (that function should work in both client and server side). Something like:
```js
// If you are running this app in subdirectory,
// we need to add basename prop for BrowserRouter and StaticRouter
const basenameMaybe = canonicalRootURL => {
  const rootURLWithoutProtocol = canonicalRootURL.indexOf('//') >= 0
    ? canonicalRootURL.split('//')[1]
    : canonicalRootURL;
  const pathParts = rootURLWithoutProtocol.split('/');

  if (pathParts.length > 1 && pathParts[1].length > 0) {
    const path = pathParts.filter((p, index) => index > 0).join('/');
    return { basename: `/${path}` };
  }
  return {};
};
```

It can be then used on client side as:
```js
<BrowserRouter {...basenameMaybe(config.canonicalRootURL)}>
```
and on SSR:
```js
<StaticRouter location={url} context={context} {...basenameMaybe(config.canonicalRootURL)}>
```

**4) Update changelog**
`- [add] <some description> [#1058](https://github.com/sharetribe/flex-template-web/pull/1058)`

And remember to run `yarn run format` & `yarn run format-docs`

lastlink:
This was mostly for building the static content not for server side, but your solution may be better. I found `PUBLIC_URL` to be needed since your building tools are a fork off of create react app and this was also needed to correct some of the build files in the `sharetribe-scripts` that weren't being added properly such as the generated [hash].js and [hash].css files by giving them abosulte paths

Gnito:
I haven't had time to check if there are problems with "PUBLIC_URL" (or not using it). I'm closing this now and add the task to our feature request backlog. 

We can reopen this if you want to continue with the pull request.

PR #1059, opened 2019-03-30
Patch 1

lastlink:
* support for code to currency

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -29,7 +29,7 @@ const sortSearchByDistance = false;
 //
 // In a way, 'processAlias' defines which transaction process (or processes)
 // this particular web application is able to handle.
-const bookingProcessAlias = 'preauth-with-nightly-booking/release-1';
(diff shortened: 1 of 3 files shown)


Gnito:
Hi @lastlink, is this intentional pull request or an accidental one?

I'm not going to accept this one since it's breaking quite many rules we have internally used for template app - and essentially this fits better into the code base of a customization project.

Some rules we try to follow internally:
- Translation keys should always point to a component (makes it easier for translators to understand where it is used).
- _Avoid_ dynamic translation keys if possible since it makes it harder to find a correct file/line by searching with a translation key.
- Don't use string concatenation (It might cause problems with translations/localization).

And pushing basic configurations to environment variables makes it unnecessarily complex. In this case,  changing variables like `bookingProcessAlias ` and `bookingUnitType` are more than likely to cause major customization needs (since most marketplaces have their own transaction process). So, those are not something that can be carelessly swapped in production.

PR #1062, opened 2019-04-03
Add support for arbitrary line item codes

lyyder:
**Update 2019-04-05: The booking breakdown layout has been improved for a more visible sub total row, see updated screenshots**

With custom pricing line items can have any sort of code as long as it begins with `line-item/`.

This PR loosens the line item code validation and changes booking breakdown so that all unknown line items are rendered too.

About rendering _unknown_ line items, __initial plan__ was to render quantity based line items on top of subtotal and the percentage based items below it:

```
Car wash        $20
Car cleaning    $10
Sub total       $30
Discount       -$5
Commission     $4
---
Total           $29
```

However, this design got a bit tricky with refunds. Previously the refund has been right after sub total and commissions have their own refund line. If percentage based unknown line items where below the sub total, should they have their own refund line in the breakdown. The simplest solution was to add everything except commissions above the sub total, refund matching those lines below sub total and commissions as they are. The main upside of this design is that the refund always matches the sub total which makes it clear to read.

So current the current design is:

Screenshot 2019-04-05 at 14 40 11

And with refund:

Screenshot 2019-04-05 at 14 40 17

The __downside__ of the design is that unknown line items only present the human readable code and line total. If a line item has quantity other than one for example, it would be nice to be able to read from the breakdown. Maybe one way to show it would be:

```
2 x Car wash ($20)      $40
```


src/util/types.js
```diff
@@ -323,7 +323,13 @@ propTypes.transaction = shape({
 
     lineItems: arrayOf(
       shape({
-        code: oneOf(LINE_ITEMS).isRequired,
+        code: (props, propName, componentName) => {
```
Gnito: Could you extract this anonymous function as 
```js
const requiredLineItemPropType = (props, propName, componentName) => { /*...*/};
// or lineItemPropTypeIsRequired
```
```js
code: requiredLineItemPropType,
```

src/util/data.js
```diff
@@ -333,3 +333,21 @@ export const overrideArrays = (objValue, srcValue, key, object, source, stack) =
...
+ *
+ * @return {string} returns the line item code humanized
+ */
+export const humanizeLineItemCode = code => {
+  if (!/^line-item\/.+/.test(code)) {
+    throw new Error(`Invalid line item code: ${code}`);
+  }
+  const lowercase = code.replace(/^line-item\//, '').replace(/-/g, ' ');
+
+  return lowercase.charAt(0).toUpperCase() + lowercase.slice(1);
```
Gnito: Btw. if clients can create line-items, can they create codes with format like `line-item/my-Banana-Inc-fee` and `line-item/myBananaIncFee`.

There's probably no need at this point address everything - I'm just not sure what is accepted as custom line-item by API?

src/components/BookingBreakdown/LineItemsUnknownMaybe.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+import React from 'react';
+import { intlShape } from 'react-intl';
+import { formatMoney } from '../../util/currency';
+import { humanizeLineItemCode } from '../../util/data';
+import { LINE_ITEMS, propTypes } from '../../util/types';
+
+import css from './BookingBreakdown.css';
+
+const LineItemsUnknownMaybe = props => {
```
Gnito: Name suggestion: `LineItemUnknownItemsMaybe`

src/components/BookingBreakdown/LineItemRefundMaybe.js
```diff
@@ -1,23 +1,62 @@
...
+ */
+const nonCommissionReversalLineItems = transaction => {
+  return transaction.attributes.lineItems.filter(item => !isCommission(item) && item.reversal);
+};
+
+const LineItemRefundMaybe = props => {
+  const { transaction, intl } = props;
+
+  // all non-commission, reversal line items
+  const lineItems = nonCommissionReversalLineItems(transaction);
```
Gnito: Maybe this could be renamed as `const refundLineItems = `?

src/util/data.js
lyyder: @Gnito the API takes anything up to 64 chars that starts with _line-item/_, so it can be myBananaIncFee too. Good remark, I could add a line item code format suggestion to the how-to guide.

src/components/BookingBreakdown/LineItemRefundMaybe.js
lyyder: @Gnito This follows the LineItemSubTotalMaybe component in naming the line items const but maybe I could rename in both.

PR #1077, opened 2019-05-06
Fix error message on logout with Sentry

ggdev:
Current implementation causes an error popup to be displayed on logout if Sentry is enabled. This is because `remove_user` is not a valid method on the `scope` object. Changing to use the `setUser` with a `null` parameter as discussed here - getsentry/sentry-docs#500


OtterleyW:
I need to test this and for that, I created another PR [#1080](https://github.com/sharetribe/flex-template-web/pull/1080). Thanks for spotting this!

PR #1078, opened 2019-05-06
Fix spurious missing payment info modal

ggdev:
Current check for Stripe account setup is looking for non-existent `stripeAccount` property, causing a spurious modal asking to set up payment info even though it has already been configured. Changing to check the correct property on the `currentUser` object.


Gnito:
@ggdev could you elaborate a bit more about how to replicate this error. We have moved away from `attributes.stripeConnect` on purpose and instead check the relatively new entity called `stripeAccount`. This is related to breaking change in Stripe API (`2019-02-19`). At that point, we had to create new API endpoints and entities. 

Currently, we include `stripeAccount` entity always to the `fetchCurrentUser` call (in src/ducks/user.duck.js). So, I'm not sure why that nagging modal would fire at the wrong time.

ggdev:
Steps to reproduce:

Pre-req: Register user, save bank payment details, and logout

1. Login
2. Navigate to "Profile Settings" via nav menu
3. Change your Bio and "Save Changes"
4. Receive error modal

Added alternate fix by including `stripeAccount` in `updateProfile` API call

Gnito:
@ggdev thanks for raising this up! It actually created a bit of discussion on our side.

Previously, we were almost leaning a bit towards deprecating `stripeConnected` attribute, but since that would mean that `stripeAccount` entity needs to be included every time an update or show query is made for currentUser,  we ended up reversing that change. (i.e. ProfileSettingsPage, ProfilePage, ContactDetailsPage, etc.)

I picked your commit and created another PR to test it:
https://github.com/sharetribe/flex-template-web/pull/1084

PR #1079, opened 2019-05-07
Don't mutate args in updatedEntities

OtterleyW:


src/util/data.js
```diff
@@ -52,7 +53,7 @@ export const updatedEntities = (oldEntities, apiResponse) => {
     const entity = entities[type][id.uuid];
```
Gnito: Deep cloning might be too heavy, but spread, at least, this entity and its attributes (can be done on line 29).

PR #1081, opened 2019-05-07
Testing CRA 3 update

Gnito:
Testing how CRA v3.0.0 works with FTW. Check also [changelog from CRA fork](https://github.com/sharetribe/create-react-app/blob/master/CHANGELOG.md)

- Reserve use* function naming pattern for React Hooks. (For some reason, Rules of Hooks eslint plugin didn't like our `useInitialValues` function (so, we decided to rename it and reserve use* function names for Hooks)

- Recent SDK update changed the proptypes for snapshots.
- Update scripts/config.js after Prettier version bump
- Remove unnecessary audit exceptions


package.json
```diff
@@ -50,21 +50,25 @@
...
     "seedrandom": "^2.4.4",
     "sharetribe-flex-sdk": "^1.3.0",
-    "sharetribe-scripts": "2.1.8",
+    "sharetribe-scripts": "3.0.0-rc",
     "smoothscroll-polyfill": "^0.4.0",
     "source-map-support": "^0.5.9",
     "url": "^0.11.0"
   },
   "devDependencies": {
+    "babel-jest": "24.7.1",
```
Gnito: There was a strange transient dependency problem without this. So, I had to include it as explicit dev-dependency:
https://github.com/timarney/react-app-rewired/issues/388

PR #1088, opened 2019-05-13
Use final form in StripePaymentForm

OtterleyW:


src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -539,8 +541,8 @@ CheckoutPageComponent.propTypes = {
   }).isRequired,
   sendOrderRequest: func.isRequired,
   onCreateStripePaymentToken: func.isRequired,
-  stripePaymentTokenInProgress: bool.isRequired,
-  stripePaymentTokenError: bool.isRequired,
+  stripePaymentTokenInProgress: bool,
+  stripePaymentTokenError: bool,
```
Gnito: Use actual error object (and validate propTypes with our own `propTypes.error` from `../../util/types`)

src/forms/StripePaymentForm/StripePaymentForm.js
```diff
@@ -1,9 +1,16 @@
...
+ * Card is not a Final Form field so it's not available trough Final Form.
+ * It's also handled separately in handleSubmit function.
+ */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
+import { Form as FinalForm } from 'react-final-form';
 import classNames from 'classnames';
-import { Form, PrimaryButton, ExpandingTextarea } from '../../components';
+import { Form, PrimaryButton, FieldTextInput } from '../../components';
```
Gnito: Reorder these so that components are after config and util files:
// external libs
// config & utils
// shared components

// css & local subcomponents from this directory

src/ducks/stripe.duck.js
```diff
@@ -509,7 +509,7 @@ export const createStripePaymentToken = params => dispatch => {
     })
     .catch(err => {
       const e = storableError(err);
-      dispatch(stripeAccountCreateError(e));
+      dispatch(stripePaymentTokenCreateError(e));
```
Gnito: One more thing: 
since this thunk function is not talking with our API/SDK but Stripe's SDK instead - `apiErrors` is not relevant. Just use "message" key from given [Error object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).

Gnito: [apiErrors](https://github.com/sharetribe/flex-template-web/blob/master/src/util/errors.js#L219) is just our mapping on top of [errors array](https://www.sharetribe.com/api-reference/#errors) given by Marketplace API.

Gnito: You could also shorten some variables and function names since these are inside stripe.duck.js file (no need to repeat "stripe" word).
```js
//export const STRIPE_PAYMENT_TOKEN_CREATE_REQUEST = 'app/stripe/STRIPE_PAYMENT_TOKEN_CREATE_REQUEST';
export const CREATE_PAYMENT_TOKEN_REQUEST = 'app/stripe/CREATE_PAYMENT_TOKEN_REQUEST';
```
```js
// export const stripePaymentTokenCreateRequest = () => ({
export const createPaymentTokenRequest = () => ({
```

Here's an example of how I added `stripe.handleCardPayment` thunk call:
https://github.com/sharetribe/flex-template-web/pull/1090/files

PR #1091, opened 2019-05-16
Use handleCardPayment thunk

Gnito:
This is still missing Marketplace API thunk calls.
(Assumed names for those are `requestPayment` & `confirmPayment`)

### DIFF
--- a/src/containers/CheckoutPage/CheckoutPage.js
+++ b/src/containers/CheckoutPage/CheckoutPage.js
@@ -9,7 +9,13 @@ import config from '../../config';
 import routeConfiguration from '../../routeConfiguration';
 import { pathByRouteName, findRouteByRouteName } from '../../util/routes';
 import { propTypes, LINE_ITEM_NIGHT, LINE_ITEM_DAY } from '../../util/types';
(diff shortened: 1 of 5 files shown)


src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -138,61 +153,122 @@ export class CheckoutPageComponent extends Component {
...
+        ? Promise.resolve(paymentIntent)
+        : onHandleCardPayment(params);
+    };
+
+    const fnConfirmPayment = parameters => {
+      const { message } = handlePaymentParams;
+      const params = { ...parameters, message };
+      return Promise.resolve(params); // onConfirmPayment(params);
+    };
+
```
OtterleyW: Adding a comment to clarify what magic is happening here and why might be useful.

src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -138,61 +153,122 @@ export class CheckoutPageComponent extends Component {
...
+    const applyAsync = (acc, val) => acc.then(val);
```
OtterleyW: Adding a comment to clarify what magic is happening here and why might be useful.

PR #1104, opened 2019-05-27
prevent error when a provider commission does not exist

rdoh:


src/components/BookingBreakdown/LineItemProviderCommissionMaybe.js
```diff
@@ -12,11 +12,7 @@ const { Money } = sdkTypes;
 // Validate the assumption that the commission exists and the amount
 // is zero or negative.
 const isValidCommission = commissionLineItem => {
-  return (
-    commissionLineItem &&
-    commissionLineItem.lineTotal instanceof Money &&
-    commissionLineItem.lineTotal.amount <= 0
-  );
+  return commissionLineItem.lineTotal instanceof Money && commissionLineItem.lineTotal.amount <= 0;
```
OtterleyW: We should check if the `commissionLineItem` exists before trying to access `lineTotal` or `amount`. It's not clear why this change has been made.

OtterleyW:
Added just the check for `providerCommissionLineItem` in PR https://github.com/sharetribe/flex-template-web/pull/1112

PR #1106, opened 2019-05-31
Fix bug in showing the review link after other party has already reviewed it

OtterleyW:
Because of the bug, the second review link was not visible
  in `ActivityFeed`. Instead, the link to the first review was shown even if the user had already sent the review. 

This bug was introduced most likely in FTW v2.10.0 when the `ActivityFeed` was refactored.

### DIFF
--- a/src/components/ActivityFeed/ActivityFeed.js
+++ b/src/components/ActivityFeed/ActivityFeed.js
@@ -25,6 +25,7 @@ import {
   isCustomerReview,
   isProviderReview,
   txRoleIsProvider,
+  txRoleIsCustomer,
   getUserTxRole,
   isRelevantPastTransition,
 } from '../../util/transaction';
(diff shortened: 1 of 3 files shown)


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -152,7 +153,7 @@ const resolveTransitionMessage = (
       const userHasLeftAReview = hasUserLeftAReviewFirst(ownRole, transaction);
 
       const reviewLink =
-        reviewPeriodJustStarted || !(reviewPeriodIsOver || userHasLeftAReview) ? (
+        reviewPeriodJustStarted && !(reviewPeriodIsOver || userHasLeftAReview) ? (
```
Gnito: I don’t understand this change.

If the user is reviewing as a second person:
false && !(false || false)
false && true 
false

It’s probably better to refactor this a bit:

```js
const waitingForMissingSecondReview = !(reviewPeriodIsOver || userHasLeftAReview);
const reviewLink = reviewPeriodJustStarted || waitingForMissingSecondReview ? (
```

OtterleyW: This part is only for the first review. The second review is handled in a different case (but it seems there was a mistake in the code comment). 

So actually checking `!(reviewPeriodIsOver || userHasLeftAReview)` is not needed here at all because it's enough to check that the current state is "delivered".

PR #1109, opened 2019-06-04
WiP: Confirm payment

OtterleyW:
Confirm the order by transitioning to "confirm-payment"

### DIFF
--- a/src/containers/CheckoutPage/CheckoutPage.js
+++ b/src/containers/CheckoutPage/CheckoutPage.js
@@ -41,7 +41,12 @@ import { StripePaymentForm } from '../../forms';
 import { isScrollingDisabled } from '../../ducks/UI.duck';
 import { handleCardPayment } from '../../ducks/stripe.duck.js';
 
-import { initiateOrder, setInitialValues, speculateTransaction } from './CheckoutPage.duck';
+import {
+  initiateOrder,
+  setInitialValues,
+  speculateTransaction,
+  confirmPayment,
+} from './CheckoutPage.duck';
 import { storeData, storedData, clearData } from './CheckoutPageSessionHelpers';
 import css from './CheckoutPage.css';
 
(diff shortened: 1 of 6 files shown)


src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -153,26 +159,27 @@ export class CheckoutPageComponent extends Component {
...
     const fnRequestPayment = fnParams => {
       // fnParams should be { listingId, bookingStart, bookingEnd }
-      const {
-        pageData,
-        stripePaymentIntentClientSecret,
-        stripePaymentIntentId,
-      } = handlePaymentParams;
+      const { pageData } = handlePaymentParams;
+
+      const { stripePaymentIntentClientSecret, stripePaymentIntentId } = fnParams;
```
Gnito: This is step 1, so these stripePaymentIntentClientSecret and stripePaymentIntentId are not actually available unless user has completed `onInitiateOrder` call previously. My previous code was expecting clientSecret to be saved in sessionStorage, but it will likely be saved among transaction entity.

Just remove this line and in line 175 return `? Promise.resolve(pageData.enquiredTransaction)`. 
(Although, I'll change that line in my PR.)

src/ducks/stripe.duck.js
```diff
@@ -538,7 +538,10 @@ export const createStripeAccount = payoutDetails => (dispatch, getState, sdk) =>
 export const handleCardPayment = params => dispatch => {
   // It's required to use the same instance of Stripe as where the card has been created
   // so that's why Stripe needs to be passed here and we can't create a new instance.
-  const { stripe, card, stripePaymentIntentClientSecret, paymentParams } = params;
+  const { stripe, card, paymentParams, id } = params;
+  const {
+    stripePaymentIntentClientSecret,
+  } = params.attributes.protectedData.stripePaymentIntents.default;
```
Gnito: I think it's better to simplify these params, before this `handleCardPayment` function is called.

```js
const { stripe, card, stripePaymentIntentClientSecret, paymentParams, orderId } = params;
```

This could be done in CheckoutPage.js line 185 (fnParams). `initiateOrder` is currently returning a full transaction entity - so, the comment above that line is wrong:
```js
const fnHandleCardPayment = fnParams => {
      // fnParams should be { stripePaymentIntentClientSecret, stripePaymentIntentId }
```
It should be something like
```js
const fnHandleCardPayment = fnParams => {
      // fnParams should be returned transaction entity
      const order = fnParams;
      const stripePaymentIntentClientSecret = order.attributes.protectedData.stripePaymentIntents.default;
      const orderId = order.id;

      const { stripe, card, billingDetails, paymentIntent } = handlePaymentParams;
      const params = {
        stripePaymentIntentClientSecret,
        orderId,
       // other params that previously were below "...fnParams"
```

src/ducks/stripe.duck.js
```diff
@@ -549,7 +552,8 @@ export const handleCardPayment = params => dispatch => {
         return Promise.reject(response);
       } else {
         dispatch(handleCardPaymentSuccess(response));
-        return response;
+        const values = { ...response, id };
+        return values;
```
Gnito: How about just `return { ...response, orderId };`?
(My suggestion in previous comment was to rename "id" to "orderId" - but it can be "transactionId" too)

src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -208,18 +220,21 @@ export class CheckoutPageComponent extends Component {
     const handlePaymentIntentCreation = composeAsync(
       fnRequestPayment,
       fnHandleCardPayment,
-      fnConfirmPayment
+      fnConfirmPayment,
+      fnSendInitialMessage
```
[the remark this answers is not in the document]
Gnito: Btw. in `master` branch, we have a bit artificial restriction that message can be sent from `initiateOrder` call, but not from `initiateOrderAfterEnquiry`. So, this restriction is now disappearing, but personally, I think that it is just a good change.
https://github.com/sharetribe/flex-template-web/blob/master/src/containers/CheckoutPage/CheckoutPage.duck.js#L171

src/ducks/stripe.duck.js
Gnito: Btw. we probably should not assume that `protectedData.stripePaymentIntents.default` always exists. Our clients might make strange TX process graphs - and even if they don't make, they might take an update from upstream before their current tx process supports PaymentIntents. 

This dot-chain will throw an exception if `protectedData` or `protectedData.stripePaymentIntents` is missing. Throwing an error might be a correct thing to do, but we should throw it explicitly with a better error message.

PR #1112, opened 2019-06-10
Prevent provider commission error

OtterleyW:


OtterleyW:
Added just the check for `providerCommissionLineItem` in PR https://github.com/sharetribe/flex-template-web/pull/1112

PR #1114, opened 2019-06-11
Clear cardtoken before redirecting

OtterleyW:
`stripeCardToken` didn't update when the user tried to book the same listing for a second
  time. This update will clear the old card token from Redux store when redirecting to
  `TransactionPage`.

### DIFF
--- a/src/containers/CheckoutPage/CheckoutPage.js
+++ b/src/containers/CheckoutPage/CheckoutPage.js
@@ -38,7 +38,7 @@ import {
   setInitialValues,
   speculateTransaction,
 } from './CheckoutPage.duck';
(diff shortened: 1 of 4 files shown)


src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -190,6 +190,7 @@ export class CheckoutPageComponent extends Component {
         const orderDetailsPath = pathByRouteName('OrderDetailsPage', routes, {
           id: orderId.uuid,
         });
+        this.props.onClearStripePaymentToken();
```
Gnito: I'd spread this out on line 150 among other props.

PR #1121, opened 2019-07-03
Verify email automatically and redirect to landing page

OtterleyW:


src/containers/EmailVerificationPage/EmailVerificationPage.js
```diff
@@ -57,7 +58,13 @@ export const EmailVerificationPageComponent = props => {
     id: 'EmailVerificationPage.title',
   });
 
-  const initialValues = { verificationToken: parseVerificationToken(location) };
+  const initialValues = {
+    verificationToken: parseVerificationToken(location ? location.search : null),
+  };
+
+  if (currentUser && currentUser.attributes && currentUser.attributes.emailVerified) {
```
Gnito: This works and is pretty clear, but I'd like us to use `ensureCurrentUser` more often (from util/data).

```suggestion
  const user = ensureCurrentUser(currentUser);
  if (user && user.attributes.emailVerified) {
```
`const user` or `const ensuredCurrentUser`

Then we need to import the ensureCurrentUser and change 
- line ~71 aka `{currentUser ? (` to `{user.id ? (`
- line ~75 aka `currentUser={currentUser}` to `currentUser={user}`

PR #1129, opened 2019-07-05
Add keyword search

OtterleyW:
Add keyword search components to `SearchFilters.js` and `SearchFiltersMobile.js`. On mobile live edit is in use so we need to use debounce in `FilterForm.js` so that the query is made after the user has stopped typing.

**On desktop:**
keyword-search-desktop

**On mobile:**
keyword-search-mobile


src/containers/SearchPage/SearchPage.helpers.js
```diff
@@ -13,11 +13,12 @@ import routeConfiguration from '../../routeConfiguration';
  * @param {Object} paramValue Search parameter value
  * @param {Object} filters Filters configuration
  */
-export const validURLParamForExtendedData = (paramName, paramValue, filters) => {
+export const validURLParamForExtendedData = (paramName, paramValueRaw, filters) => {
   const filtersArray = Object.values(filters);
   // resolve configuration for this filter
   const filterConfig = filtersArray.find(f => f.paramName === paramName);
 
+  const paramValue = paramValueRaw.toString();
```
OtterleyW: I added this because if the value in keyword search is e.g. "123" the app would crash otherwise as the value is a number and not a string and it doesn't have split() function.

src/forms/FilterForm/FilterForm.js
```diff
@@ -68,7 +81,12 @@ const FilterFormComponent = props => {
           >
             <div className={classNames(paddingClasses || css.contentWrapper)}>{children}</div>
 
-            {liveEdit ? (
+            {useDebounce && liveEdit ? (
```
OtterleyW: `KeywordFilter` is using our generic filter component so I added an option to use debounce in the generic `FilterForm`.  I don't know if it's ok or should I create new specific form etc. for `KeywordFilter`.

Gnito: I moved this debounce to KeywordFilter component

PR #1134, opened 2019-07-09
Create payment methods page

OtterleyW:
localhost_3000_account_payment-methods


src/components/StripePaymentAddress/StripePaymentAddress.css
```diff
@@ -0,0 +1,30 @@
+@import '../../marketplace.css';
+
+.root {
+}
+
+.paymentAddressField {
+  padding-top: 38px;
+}
+
+.formRow {
```
Gnito: I'd recommend that you try to keep these rules in some kind of order.

For example, _flex-shrink_ is something that expects parent component to use flexbox layout. => it would be better to have those rules first.
https://github.com/sharetribe/flex-template-web/blob/c30a6ab6e22e1bd6f4c155d435a0ea485b76c33c/src/components/LayoutSideNavigation/LayoutSideNavigation.css#L11-L18

Btw. `.sectionContainer` is used in the code, but there's no rule for it. (maybe it should refer to `.root`?)

src/containers/PaymentSettingsPage/PaymentSettingsPage.js
```diff
@@ -0,0 +1,138 @@
...
+  LayoutWrapperTopbar,
+  LayoutWrapperFooter,
+  Footer,
+  Page,
+  UserNav,
+} from '../../components';
+import { TopbarContainer } from '../../containers';
+import { PaymentSettingsForm } from '../../forms';
+
+import css from './PaymentSettingsPage.css';
```
Gnito: The page component look good 👍 
I'd want you to make one big renaming task. In design, this page is called Payment Methods - so, it's probably better that the component, route, title, and design use the same name.

Screenshot 2019-07-18 at 12 42 42

src/containers/PaymentSettingsPage/PaymentSettingsPage.duck.js
```diff
@@ -0,0 +1 @@
+// TODO
```
Gnito: At some point, this should contain a call `createStripeSetupIntent`

I was also thinking that a fetch for currentUser with a relationship: "stripeCustomer.defaultPaymentMethod" should be here. But maybe just a "wrapping" thunk function:

fetchCurrentUserWithPaymentMethod:
1) request: set state: fetchingPaymentMethod: 'inProgress'
=> user.duck.js :: fetchCurrentUser({relationship})
// fetchCurrentUser needs to be modified so that it can accept (optional) params object.
2) success: fetchingPaymentMethod: "succeeded"
3) error: fetchingPaymentMethod: "error"

Note: since currentUser is already fetched, we need to check if fetchCurrentUser is called with relationship stripeCustomer.defaultPaymentMethod.
 (Another option is to always fetch currentUser with the aforementioned relationship, but then it fetches unnecessary data ~90% of time.)

src/routeConfiguration.js
```diff
@@ -249,6 +250,14 @@ const routeConfiguration = () => {
       component: props => <PayoutPreferencesPage {...props} />,
       loadData: PayoutPreferencesPage.loadData,
     },
+    {
+      path: '/account/payment-settings',
```
Gnito:
```suggestion
      path: '/account/payment-methods',
```

PR #1142, opened 2019-07-23
user.duck.js can fetch currentUser with different relationships.

Gnito:
`user.duck.js` can fetch currentUser with different relationships.
It merges old and new relationships.

After this change, we could call `fetchCurrentUser` function as:
```js
fetchCurrentUser({ include: ['stripeCustomer.defaultPaymentMethod'] });
```

### DIFF
--- a/src/ducks/user.duck.js
+++ b/src/ducks/user.duck.js
@@ -40,6 +40,20 @@ export const SEND_VERIFICATION_EMAIL_ERROR = 'app/user/SEND_VERIFICATION_EMAIL_E
 
 // ================ Reducer ================ //
 
+const mergeCurrentUser = (oldCurrentUser, newCurrentUser) => {
+  const { id: oId, type: oType, attributes: oAttr, ...oldRelationships } = oldCurrentUser || {};
+  const { id, type, attributes, ...relationships } = newCurrentUser || {};
+
+  // Passing null will remove currentUser entity.
+  // Only relationships are merged.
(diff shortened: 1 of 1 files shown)


src/ducks/user.duck.js
```diff
@@ -301,13 +315,13 @@ export const fetchCurrentUser = () => (dispatch, getState, sdk) => {
     return Promise.resolve({});
   }
 
-  const params = {
+  const parameters = params || {
```
OtterleyW: Did I understand this correctly that by default `profileImage` and `stripeAccount` are included but if a different relationship is passed to the `fetchCurrentUser` then these defaults are not included?

Gnito: @OtterleyW yes. I was thinking of creating some kind of extra params pattern, but since currentUser is always fetched if the user is authenticated, I decided that `mergeCurrentUser` function will handle those cases. `mergeCurrentUser` is needed anyway, otherwise, someone would need to keep track that which relationships need to be included by different pages.

So, `fetchCurrentUser({ include: ['stripeCustomer.defaultPaymentMethod'] });` will add `stripeCustomer` to the previously fetched `currentUser` entity.

If there's a need to ensure that profileImage and stripeAccount are included for the page component, then it's better to call all of them 
```js
fetchCurrentUser({ 
  include: ['stripeCustomer.defaultPaymentMethod', 'profileImage', 'stripeAccount'],
  'fields.image': ['variants.square-small', 'variants.square-small2x'] });
```

PR #1144, opened 2019-07-24
Update duck files

OtterleyW:
- Add `handleCardSetup` to `stripe.duck.js`
- Add `createStripeCustomer` to `paymentMethods.duck.js`

### DIFF
--- a/src/ducks/paymentMethods.duck.js
+++ b/src/ducks/paymentMethods.duck.js
@@ -1,2 +1,182 @@
-// This file will contain Flex API calls to handle saving payment methods.
-// CheckoutPage and PaymentMethodsPage will share these thunk calls.
+import { storableError } from '../util/errors';
+import * as log from '../util/log';
+
+// ================ Action types ================ //
+
+export const CREATE_STRIPE_CUSTOMER_REQUEST = 'app/stripe/CREATE_STRIPE_CUSTOMER_REQUEST';
+export const CREATE_STRIPE_CUSTOMER_SUCCESS = 'app/stripe/CREATE_STRIPE_CUSTOMER_SUCCESS';
+export const CREATE_STRIPE_CUSTOMER_ERROR = 'app/stripe/CREATE_STRIPE_CUSTOMER_ERROR';
+
+export const ADD_PAYMENT_METHOD_REQUEST = 'app/stripe/ADD_PAYMENT_METHOD_REQUEST';
+export const ADD_PAYMENT_METHOD_SUCCESS = 'app/stripe/ADD_PAYMENT_METHOD_SUCCESS';
+export const ADD_PAYMENT_METHOD_ERROR = 'app/stripe/ADD_PAYMENT_METHOD_ERROR';
+
+export const DELETE_PAYMENT_METHOD_REQUEST = 'app/stripe/DELETE_PAYMENT_METHOD_REQUEST';
+export const DELETE_PAYMENT_METHOD_SUCCESS = 'app/stripe/DELETE_PAYMENT_METHOD_SUCCESS';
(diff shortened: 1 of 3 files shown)


src/ducks/stripe.duck.js
```diff
@@ -625,3 +659,48 @@ export const handleCardPayment = params => dispatch => {
       throw e;
     });
 };
+
+export const handleCardSetup = params => dispatch => {
+  console.log('handleCardSetup params', params);
```
Gnito: Remove

src/ducks/stripe.duck.js
```diff
@@ -625,3 +659,48 @@ export const handleCardPayment = params => dispatch => {
...
+  // It's required to use the same instance of Stripe as where the card has been created
+  // so that's why Stripe needs to be passed here and we can't create a new instance.
+  const { stripe, card, setupIntentClientSecret } = params;
+
+  dispatch(handleCardSetupRequest());
+
+  return stripe
+    .handleCardSetup(setupIntentClientSecret, card, {
+      payment_method_data: {
+        billing_details: { name: params.name },
```
Gnito: Remember to expand this so that it includes all the billing details.

src/ducks/stripe.duck.js
```diff
@@ -625,3 +659,48 @@ export const handleCardPayment = params => dispatch => {
...
+        return response;
+      }
+    })
+    .catch(err => {
+      // Unwrap Stripe error.
+      const e = err.error || storableError(err);
+      dispatch(handleCardSetupError(e));
+
+      // Log error
+      const containsSetupIntent = err.error && err.error.payment_intent;
```
Gnito: Is that right?
`containsSetupIntent` vs `err.error.payment_intent`

Gnito: Also rows below reference `payment_intent`.

src/ducks/paymentMethods.duck.js
```diff
@@ -1,2 +1,182 @@
...
+  dispatch(stripeCustomerCreateRequest());
+  return sdk.stripeCustomer
+    .create({ stripePaymentMethodId }, { expand: true, include: ['defaultPaymentMethod'] })
+    .then(response => {
+      const stripeCustomer = response.data.data;
+      dispatch(stripeCustomerCreateSuccess(stripeCustomer));
+      return stripeCustomer;
+    })
+    .catch(e => {
+      log.error(e, 'create-stripe-user-failed');
```
Gnito: Don't log plain error. It can contain internal reference loops (Axios related loops, if I remember correctly). Those can be quite problematic if data needs to be serialized with `JSON.stringify`.

=> use `storableError(e)`

src/ducks/paymentMethods.duck.js
```diff
@@ -1,2 +1,182 @@
...
+export const deletePaymentMethodError = e => ({
+  type: DELETE_PAYMENT_METHOD_ERROR,
+  payload: e,
+  error: true,
+});
+
+// ================ Thunks ================ //
+
+export const createStripeCustomer = params => (dispatch, getState, sdk) => {
+  const stripePaymentMethodId = params.setupIntent.payment_method;
```
Gnito: This is a bit odd. Why not just pass setupIntent instead of wrapping it with some object?

```js
export const createStripeCustomer = setupIntent => (dispatch, getState, sdk) => {
  const stripePaymentMethodId = setupIntent.payment_method;
```
or even paymentMethod directly:

```js
export const createStripeCustomer = paymentMethod => (dispatch, getState, sdk) => {
  const stripePaymentMethodId = paymentMethod;
```

i.e. extract the relevant data before calling these thunk functions.

Gnito: This is a bit related to the other branch on top of this one.

PR #1145, opened 2019-07-26
Save payment method on PaymentMethodsPage

OtterleyW:
- Create Stripe setup intent
- Create Stripe customer if that doesn't exist already
- Save the default payment method
 - If there is a default payment method already saved when trying to save a new one, update payment method (delete the old one and add a new one)
- Delete the default payment method


src/containers/PaymentMethodsPage/PaymentMethodsPage.duck.js
```diff
@@ -1 +1,95 @@
-//TODO
+import { fetchCurrentUser } from '../../ducks/user.duck';
+import { storableError } from '../../util/errors';
+import * as log from '../../util/log';
+
+// ================ Action types ================ //
+
+export const SETUP_INTENT_REQUEST = 'app/CheckoutPage/SETUP_INTENT_REQUEST';
```
Gnito: Rename these strings. 'app/PaymentMethodsPage/SETUP_INTENT_REQUEST'
(These are for debugging purposes. We can track which calls are sent to reducers and in which order.)

src/containers/PaymentMethodsPage/PaymentMethodsPage.duck.js
```diff
@@ -1 +1,95 @@
...
+// ================ Thunks ================ //
+
+export const createStripeSetupIntent = () => (dispatch, getState, sdk) => {
+  dispatch(setupIntentRequest());
+  return sdk.stripeSetupIntents
+    .create()
+    .then(response => {
+      dispatch(setupIntentSuccess());
+      const paymentIntent = response.data.data;
+      return paymentIntent;
```
Gnito: paymentIntent? setupIntent maybe.

Gnito: Is there any need to reference created setupIntent later? I.e. should it be added to store too `setupIntentSuccess(setupIntent)` ?

src/containers/PaymentMethodsPage/PaymentMethodsPage.duck.js
```diff
@@ -1 +1,95 @@
...
+export const stripeCustomer = () => (dispatch, getState, sdk) => {
+  dispatch(stripeCustomerRequest());
+
+  return dispatch(fetchCurrentUser({ include: ['stripeCustomer.defaultPaymentMethod'] }))
+    .then(response => {
+      dispatch(stripeCustomerSuccess());
+    })
+    .catch(e => {
+      log.error(storableError(e), 'fetch-stripe-customer-failed');
+      dispatch(stripeCustomerError(storableError(e)));
```
Gnito: You could reduce calls to `storableError` by saving the result to a variable.

src/containers/PaymentMethodsPage/PaymentMethodsPage.js
```diff
@@ -119,6 +200,19 @@ export const PaymentMethodsPageComponent = props => {
...
+                formId="PaymentMethodsForm"
+                initialValues={initalValuesForStripePayment}
+                onSubmit={handleSubmit}
+                handleRemovePaymentMethod={handleRemovePaymentMethod}
+                hasDefaultPaymentMethod={hasDefaultPaymentMethod}
+                addPaymentMethodError={addPaymentMethodError}
+                deletePaymentMethodError={deletePaymentMethodError}
+                createStripeCustomerError={createStripeCustomerError}
+                handleCardSetupError={handleCardSetupError}
+              />
```
Gnito: You should also handle inProgress states (e.g. `addPaymentMethodInProgress`). Form should show spinner in the submit button if the saving is in progress.

src/containers/PaymentMethodsPage/PaymentMethodsPage.js
```diff
@@ -19,11 +26,90 @@ import {
...
+          billing_details: billingDetails,
+        },
+      };
+
+      const stripeParams = { stripe, card, setupIntentClientSecret, paymentParams };
+
+      onHandleCardSetup(stripeParams)
+        .then(result => {
+          if (result.error) {
+            return;
```
Gnito: I assume that this is not needed since there's `Promise.reject` handling in onHandleCardSetup thunk function.
This could be left in place, but it would be good to have some comment explaining what happens to errors.

src/containers/PaymentMethodsPage/PaymentMethodsPage.js
```diff
@@ -19,11 +26,90 @@ import {
...
+  } = props;
+
+  const handleSubmit = params => {
+    const ensuredCurrentUser = ensureCurrentUser(currentUser);
+    const stripeCustomer = ensuredCurrentUser.stripeCustomer;
+    const hasSavedPaymentMethod = stripeCustomer ? stripeCustomer.defaultPaymentMethod : null;
+
+    const { stripe, card, formValues } = params;
+
+    onCreateSetupIntent().then(setupIntent => {
```
Gnito: If I read this correctly, you first create `onCreateSetupIntent` and then if it succeeds, you call `onHandleCardSetup `, but there's no error handling/catch block.

How about a bit more readable chaining:

```js
onCreateSetupIntent()
  .then(setupIntent => {
    const stripeParams = { 
      stripe, 
      card, 
      setupIntentClientSecret: getClientSecret(setupIntent), // new function to extract clientSecret
      paymentParams: getPaymentParams(currentUser, formValues), // new function to extract paymentParams
    };

    return onHandleCardSetup(stripeParams); // the return value will be passed as params to the next `then` block.
  })
  .then(result => {
    const newPaymentMethod = result.setupIntent.payment_method;
    // Note: stripe.handleCardSetup returns erros as successful call (200), but those are rejected in thunk functions.

    return !stripeCustomer
      ? onCreateStripeCustomer(newPaymentMethod)
      : hasSavedPaymentMethod
      ? onUpdatePaymentMethod(newPaymentMethod)
      : onAddPaymentMethod(newPaymentMethod);
  })
  .then(() => {
    // Update currentUser entity and its sub entities: stripeCustomer and defaultPaymentMethod
    fetchStripeCustomer();
  });
  .catch(error => {
    // Handle all possible errors from onCreateSetupIntent and onHandleCardSetup if needed.
    // If there's no need, it's still good practice to always add catch and then explain in the comments why it's empty.
  });
  
```

src/containers/PaymentMethodsPage/PaymentMethodsPage.js
```diff
@@ -68,18 +154,20 @@ export const PaymentMethodsPageComponent = props => {
       }`
     : null;
 
-  const initalValuesForStripePayment = { name: userName };
+  const initalValuesForStripePayment = { name: userName, country: 'FI' };
```
Gnito: Hard-coded 'FI' sounds a bit bad for the template app - unless it's in the config.js.

PR #1151, opened 2019-08-12
Use stripe config to determine US specific fields

OtterleyW:
Don't send personal id number (/ ssn last 4) or business profile to Stripe API if they are not required in `stripe-config.js`


src/forms/PayoutDetailsForm/PayoutDetailsBusinessProfile.js
```diff
@@ -28,6 +28,10 @@ const PayoutDetailsBusinessProfile = props => {
     intl.formatMessage({ id: 'PayoutDetailsForm.businessURLRequired' })
   );
 
+  // By default, all merchant category codes (MCC) are listed in the select field. You can edit the
+  // merchantCategoryCodesUS.js and remove the codes that are not relevant to your marketplace or use a hard-coded
+  // value if there is only one code you want to use.
```
Gnito: 👍

PR #1152, opened 2019-08-12
Fixes to SavedCardDetails

OtterleyW:


Gnito:
`.cardIcon` could benefit from 
```css
  display: inline-block;
  vertical-align: baseline;
```

Another thing: [.menu](https://github.com/sharetribe/flex-template-web/blob/fix-saved-card-component/src/components/SavedCardDetails/SavedCardDetails.css#L63) can't have fixed width. It won't work on responsive layout:
Screenshot 2019-08-13 at 21 37 57


So, the `width: 564px;` needs to be removed and also remove `width: 80%;` from `.savedPaymentMethod` => it's inside flexbox - so, use `margin-right: auto;` instead.

jannekoivistoinen:
@OtterleyW great work and very well done with the baselines! I fixed few things in the component, hopefully not too heavy to review.

One more thing. When user clicks the "Delete card" and the modal opens – is it possible to add the spinner to the "Remove card" button after the user has clicked it? Now nothing happens for a second or two after the user clicks the button, the spinner would be nice way to provide feedback that the request has been sent.

PR #1153, opened 2019-08-13
Save card after onetime payment

Gnito:
Otherwise ready, but this needs error handling: 
if saving onetime payment fails, we should show an error message on the transaction page.


src/components/TransactionPanel/TransactionPanel.js
```diff
@@ -318,6 +325,12 @@ export class TransactionPanelComponent extends Component {
       id: 'TransactionPanel.sendingMessageNotAllowed',
     });
 
+    const paymentMethodsPage = (
```
OtterleyW: Maybe renaming this to `paymentMethodsPageLink` would be more descriptive and make it easier to understand where this is used.

src/ducks/paymentMethods.duck.js
```diff
@@ -171,12 +168,48 @@ export const deletePaymentMethod = () => (dispatch, getState, sdk) => {
...
+
+  return savePromise
+    .then(response => {
+      const {
+        createStripeCustomerError,
+        addPaymentMethodError,
+        deletePaymentMethodError,
+      } = getState().paymentMethods;
+
+      // If are any errors, return those errors
```
OtterleyW:
```suggestion
      // If there are any errors, return those errors
```

src/ducks/stripe.duck.js
```diff
@@ -627,8 +627,12 @@ export const handleCardPayment = params => dispatch => {
 
   dispatch(handleCardPaymentRequest());
 
+  const args = card
```
OtterleyW: Might be useful to add code comment that Stripe Element is optional and it's handled here

src/forms/StripePaymentForm/StripePaymentForm.css
```diff
@@ -42,52 +42,103 @@
...
+  margin-bottom: 0px;
+  cursor: pointer;
+}
+
+.saveForLaterUseLegalInfo {
+  @apply --marketplaceH5FontStyles;
+  color: var(--matterColorAnti);
+  display: inline-block;
+  padding: 0px 0 0px 26px;
+  margin-top: -2px;
```
OtterleyW: I assume there is a good reason to use a negative margin here but I would like to know why as it's usually something we try to avoid 👀

src/forms/StripePaymentForm/StripePaymentForm.js
```diff
@@ -74,9 +82,84 @@ const cardStyles = {
...
+    formId,
+    changePaymentMethod,
+    defaultPaymentMethod,
+    handleStripeElementRef,
+    hasCardError,
+    error,
+    paymentMethod,
+    intl,
+  } = props;
+  const last4Digits = defaultPaymentMethod.attributes.card.last4Digits;
```
OtterleyW: Is it possible that there is no `attributes` or `card` in `defaultPaymentMethod`? Should it be checked or is it handled somewhere else?

src/forms/StripePaymentForm/StripePaymentForm.js
```diff
@@ -96,40 +179,81 @@ class StripePaymentForm extends Component {
...
+    this.setState({ paymentMethod: changedTo });
+  }
+
+  handleStripeElementRef(el) {
+    this.cardContainer = el;
+    if (this.stripe && el) {
+      this.initializeStripeElement(el);
+    }
+    if (this.finalFormAPI) {
+      //this.finalFormAPI.initialize();
```
OtterleyW: Should this be removed as it's commented out?

src/forms/StripePaymentForm/StripePaymentForm.js
Gnito: Excellent question. I added `    const ensuredDefaultPaymentMethod = ensurePaymentMethodCard(defaultPaymentMethod);
` to StripePaymentForm component. (I.e. I didn't repeat that on these subcomponents.)

src/forms/StripePaymentForm/StripePaymentForm.css
Gnito: @OtterleyW Well, usually these are fixable with more consistent design. 

This time adding an extra vertical space (worth one baseline) to avoid negative margin looked bad and adjusting line-height of our fonts sounded bad too (because custom font-styling per component makes it harder to change fonts - but that's also the case with negative margins. So, I chose between two evils).

PR #1157, opened 2019-08-21
Clear old errors from PaymentMethodsPage

OtterleyW:


src/containers/PaymentMethodsPage/PaymentMethodsPage.duck.js
```diff
@@ -1,4 +1,5 @@
 import { fetchCurrentUser } from '../../ducks/user.duck';
+import { setInitialValues } from '../../ducks/paymentMethods.duck';
```
Gnito: Since this is not this reducer's own function, it's probably better to rename it:
```suggestion
import { setInitialValues as setInitialValuesForPaymentMethods } from '../../ducks/paymentMethods.duck';
```

And you need to change the call too

PR #1172, opened 2019-08-26
Update react to 16.9.0

OtterleyW:
Updating `react-dom` to v16.9.0 will cause changes to the following components:

**componentWillMount:**

- `FieldBirthdayInput`
    - Setting selected value
- `TransactionPanel`
    - Checking if mobile safari
- `CheckoutPage`
    - Loading initial data
- `NotFoundPage`
    - SSR: add ´not found´flag to context to change the response status to 404

**componentWillReceiveProps:**

- `FieldBirthdayInput`
    - Changing selected value
- `DateRangeInput`
    - Update focusedInput in case a new value for it is passed in the props. This may occur if the focus is manually set to the date picker.
- `FieldDateRangeInput`
    - Update focusedInput in case a new value for it is passed in the props. This may occur if the focus is manually set to the date picker.
- `Modal`
    - use `onManageDisableScrolling` function
- `ModalInMobile`
    - Handle change if status is changing on mobile layout or it is closing (on desktop layout)
- `ModalMissingInformation`
    - use `handleMissingInformationReminder` for determining if the modal should be shown or not
- `Page`
    - Update current scroll position, if scrolling is disabled (e.g. modal is open)
- `SearchMapWithGoogleMaps`
- `SearchMapWithMapbox`
- `StripeBankAccountTokenInput`
    -  Clear the possible input values from the state if the given country or currency changes.
- `ProfileSettingsForm`
    - Upload delay is additional time window where Avatar is added to the DOM, but not yet visible (time to load image URL from srcset)
- `Routes`
    - Calling loadData after initial rendering (on client side). This makes it possible to use loadData as default client side data loading technique. However it is better to fetch data before location change to avoid "Loading data" state.


src/components/TransactionPanel/TransactionPanel.js
```diff
@@ -99,7 +99,7 @@ export class TransactionPanelComponent extends Component {
     this.scrollToMessage = this.scrollToMessage.bind(this);
   }
 
-  componentWillMount() {
+  componentDidlMount() {
```
Gnito:
```suggestion
  componentDidMount() {
```

src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -583,9 +583,31 @@ export class CheckoutPageComponent extends Component {
       !isPaymentExpired
     );
 
-    const listingTitle = currentListing.attributes.title;
+    const listingTitle = currentListing ? currentListing.attributes.title : null;
```
Gnito: Listing prop is actually ensured already on line 520 - so, there's no need to use ternary.

Gnito: @OtterleyW btw. since you decided to move this loading-layout setup, maybe it's better to move it to somewhere near line 522

src/forms/ProfileSettingsForm/ProfileSettingsForm.js
```diff
@@ -25,10 +25,10 @@ class ProfileSettingsFormComponent extends Component {
...
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(prevProps) {
     // Upload delay is additional time window where Avatar is added to the DOM,
     // but not yet visible (time to load image URL from srcset)
-    if (this.props.uploadInProgress && !nextProps.uploadInProgress) {
+    if (prevProps.uploadInProgress && !this.props.uploadInProgress) {
       this.setState({ uploadDelay: true });
       this.uploadDelayTimeoutId = window.setTimeout(() => {
```
Gnito: There's a bug in here: componentWillUnmount function should refer to this timeout.
```
  componentWillUnmount() {
    window.clearTimeout(this. uploadDelayTimeoutId);
  }
```

Gnito: Although, it is probably better to fix this in a separate PR.

Gnito: https://github.com/sharetribe/flex-template-web/pull/1185

PR #1181, opened 2019-08-28
Update intl

OtterleyW:
**More information about the changes can be found from [Upgrade guide for react-intl@3.x](https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md)**

- Proptype `intlShape` was removed so we needed to create it again. Because of this we added a new `util/reactIntl.js` file. This file is now used to wrap all the react-intl related imports.
- `addLocaleDate` function was removed and react-intl library is now relying on native Intl APIs: [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat). In order to support older browsers we needed to add `intl-pluralrules` and `intl-relativetimeformat` to `util/polyfills.js`
- Also Node must be now compiled with `full-icu` which caused changes to `start` and `test`
    scripts in `package.json`. We also needed to add a specific config for `nodemon`
- Default `textComponent`in `IntlProvider` changed to `React.Fragment` so we need to explicitly
    set `textComponent` to `span`. Otherwise all the snapshots would have changed and it might affect to UI if there is styles added to these spans generally in customization projects.

Note: `FormattedMessage` component now supports [`tagName` prop](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedmessage) and [improved rich-text formatting](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#rich-text-formatting).

### DIFF
--- a/src/util/dates.js
+++ b/src/util/dates.js
@@ -166,7 +166,10 @@ export const formatDate = (intl, todayString, d) => {
   if (!paramsValid) {
     throw new Error(`Invalid params for formatDate: (${intl}, ${todayString}, ${d})`);
   }
-  const now = moment(intl.now());
+
+  // By default we can use moment() directly but in tests we need to use a specific dates.
+  // fakeIntl used in tests contains now() function wich returns predefined date
+  const now = intl.now ? moment(intl.now()) : moment();
   const formattedTime = intl.formatTime(d);
   let formattedDate;
 
--- a/src/util/polyfills.js
+++ b/src/util/polyfills.js
@@ -33,3 +33,20 @@ if (typeof Number.isNaN === 'undefined') {
   // eslint-disable-next-line no-self-compare
   Number.isNaN = value => value !== value;
 }
+
+// To support browsers that do not have Intl.PluralRules (e.g IE11 & Safari 12-), include this polyfill in your build.
+
+if (!Intl.PluralRules) {
+  require('intl-pluralrules');
+}
+
+// To support  browsers that do not have Intl.RelativeTimeFormat (e.g IE11, Edge, Safari 12-), include this polyfill in your build along with individual CLDR data for each locale you support.
+if (!Intl.RelativeTimeFormat) {
+  require('@formatjs/intl-relativetimeformat/polyfill');
+  require('@formatjs/intl-relativetimeformat/dist/include-aliases'); // Optional, if you care about edge cases in locale resolution, e.g zh-CN -> zh-Hans-CN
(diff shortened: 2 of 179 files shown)


src/util/test-helpers.js
```diff
@@ -24,9 +23,9 @@ const testMessages = mapValues(messages, (val, key) => key);
 // store, i18n, router, etc.
 export const TestProvider = props => {
   const store = configureStore();
-  addLocaleData([...en]);
+  //addLocaleData([...en]);
```
Gnito: remove

src/util/reactIntl.js
```diff
@@ -0,0 +1,5 @@
+import { shape } from 'prop-types';
+import { FormattedMessage, FormattedDate, FormattedHTMLMessage, injectIntl } from 'react-intl';
+
+const intlShape = shape({});
```
Gnito: This needs to contains all the correct functions

src/util/dates.js
```diff
@@ -166,7 +166,8 @@ export const formatDate = (intl, todayString, d) => {
   if (!paramsValid) {
     throw new Error(`Invalid params for formatDate: (${intl}, ${todayString}, ${d})`);
   }
-  const now = moment(intl.now());
+
+  const now = intl.now ? moment(intl.now()) : moment();
```
Gnito: Add a comment about tests / fakeIntl. I.e. we don't want those tests to fail every time the test script is run through. => fakeIntl freezes time.

src/util/polyfills.js
```diff
@@ -33,3 +33,13 @@ if (typeof Number.isNaN === 'undefined') {
...
 }
+
+if (!Intl.PluralRules) {
+  require('intl-pluralrules');
+}
+
+if (!Intl.RelativeTimeFormat) {
+  require('@formatjs/intl-relativetimeformat/polyfill');
+  require('@formatjs/intl-relativetimeformat/dist/include-aliases'); // Optional, if you care about edge cases in locale resolution, e.g zh-CN -> zh-Hans-CN
+  require('@formatjs/intl-relativetimeformat/dist/locale-data/en'); // Add locale data for de
```
Gnito: Add comments on top of the require lines - and somehow highlight the locale-data line.

PR #1194, opened 2019-09-09
Show booking  date and time correctly on InboxPage

OtterleyW:
Screen Shot 2019-09-09 at 13 39 55
Screen Shot 2019-09-09 at 13 39 35
Screen Shot 2019-09-09 at 16 36 23

### DIFF
--- a/src/components/BookingTimeInfo/BookingTimeInfo.js
+++ b/src/components/BookingTimeInfo/BookingTimeInfo.js
@@ -1,0 +1,98 @@
+import React from 'react';
+import moment from 'moment';
+import { bool, oneOf } from 'prop-types';
+import classNames from 'classnames';
+import { txIsEnquired } from '../../util/transaction';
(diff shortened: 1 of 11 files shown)


src/components/BookingTimeInfo/BookingTimeInfo.js
```diff
@@ -0,0 +1,86 @@
...
+        {`${bookingStart.date}, ${bookingStart.time} - ${bookingEnd.time}`}
+      </span>
+    ) : dateType === DATE_TYPE_DATETIME ? (
+      <>
+        <span className={css.dateSection}>{`${bookingStart.dateAndTime} - `}</span>
+        <span className={css.dateSection}>{`${bookingEnd.dateAndTime}`}</span>
+      </>
+    ) : null;
+
+  return <div className={classNames(css.bookingInfo, bookingClassName)}>{dateInfo}</div>;
```
Gnito: In this case, it might be more readable to return inside `if - else if - else` -structure.

PR #1200, opened 2019-09-16
[fix bugs] current CustomOverlayView in SearchMapWithGoogleMap does not work with new React version

rush1506:
- Adding onAdd and onRemove function to improve performance
- Rewrite current draw() method to compatible with newer react version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2019-XX-XX
 
+## [v3.5.1] 2019-09-16
+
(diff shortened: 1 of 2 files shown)


Gnito:
@rush1506 thanks for the PR. This looks good to me!

I'll merge this tomorrow and make the release.

If you notice this comment before that, could you format the code with Prettier:
- `yarn run format`
- `yarn run format-docs`

Btw. usually, we make releases in a separate PR, so there's no need to [add future release version](https://github.com/sharetribe/flex-template-web/pull/1200/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR17) within feature/bugfix pull-requests.

----

Anyway, our plan is to remove `react-google-maps` from FTW at some point since it's essentially a dead dependency. 

There's a good alternative mentioned in [the issue thread you linked](https://github.com/tomchentw/react-google-maps/issues/817): `@react-google-maps/api`. However, it would probably be better to avoid these extra dependencies and go to the same direction what we used with Mapbox (i.e. directly play with raw API within our own components, since we just need to use `google.maps.OverlayView()`).

PR #1231, opened 2019-11-22
Update API reference docs links

lyyder:
Update API reference docs links according to updated docs structure.

To be merged when the API reference change has been deployed.


Gnito:
I almost merged already - but I will wait :)

PR #1232, opened 2019-11-25
Show states and provinces as dropdown on PayoutDetailsForm

OtterleyW:


src/forms/PayoutDetailsForm/PayoutDetailsAddress.js
```diff
@@ -148,37 +134,62 @@ const PayoutDetailsAddress = props => {
...
+          disabled={disabled}
+          className={css.selectCountry}
+          autoComplete="address-level1"
+          label={provinceLabel}
+          validate={provinceRequired}
+        >
+          <option disabled value="">
+            {provincePlaceholder}
+          </option>
+          {CA_PROVINCES.map(p => (
```
Gnito: Am I correct that this array is the only difference between these different selects?
Howabout checking 

```js
const states = showStateUS 
  ? US_STATES 
  : showProvinceCA 
  ? CA_PROVINCES
  :  showStateAU
  ? AU_STATES
  : [];
```
And then creating one select that uses `states` array as the source of data?

PR #1233, opened 2019-11-26
Hot fix replace '<' with unicode equivalent from script tag

OtterleyW:
Fix XSS-vulnerability on `SearchPage` where URL param 'address' was exposed directly to schema, which is just a script tag: `<script type="application/ld+json">`. On server-side, this could leak malformed HTML through to browsers and made it possible to inject own script tags.

However, CSP prevents any data breach: injected JS can't send data to unknonwn 3rd party sites.

**NOTE:** Check that `REACT_APP_CSP` is in block mode on your production environment. You can read more from Flex Docs: https://www.sharetribe.com/docs/guides/how-to-set-up-csp-for-ftw/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2019-XX-XX
 
+## [v3.6.1] 2019-11-04
+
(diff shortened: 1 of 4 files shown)


himynameisubik:
Was highlighted to us (adventurent) by Nicolas Surribas (https://www.openbugbounty.org/researchers/devl00p/) via Open Bug Bounty. 

Thanks for applying the fix to the ftw-template! :)

PR #1234, opened 2019-11-28
Stripe Connect Onboarding - base branch

OtterleyW:
# Use Stripe Connect onboarding flow in FTW

In this update, we have taken into use Stripe's [Connect onboarding](https://stripe.com/docs/connect/connect-onboarding). Through the Connect onboarding Stripe will securely collect sensitive personal details and ID documents needed for verification, and handle updates to the UI to meet changing compliance requirements. On FTW we need to handle creating new Stripe accounts and saving information about the external accounts (e.g. IBAN number) to the user's Stripe account.

After this update users can also update most of the information through the Connect onboarding. Updating the external account is also possible and done through the Flex API directly.

You can read more form [How to handle provider onboarding and identity verification on FTW](https://www.sharetribe.com/docs/guides/provider-onboarding-and-identity-verification/) guide.

## Changes:

- New shared duck file `stripeConnectAccount.duck.js` which handles creating and updating the Stripe accounts and creating the Account Link where the users get redirected
- New `StripePayoutPage` which replaces the old `PayoutPreferencesPage`
- New `StripeConnectAccountForm` which replaces the old `PayoutDetailsForm` on `StripePayoutPage` and on `EditListingPage`

On the `EditListingPage`, the modal with `StripeConnectAccountForm` is shown if the user doesn't have a Stripe account yet or if there is some information missing from the account. The modal will be shown only if the user is publishing the listing so currently, users can update already published listing even if their Stripe account is restricted.

### Removed components

Because of this change, we have deprecated the old `PayoutDetailsForm` and `PayoutPreferencesPage`. Also functions related to deprecated form has been removed from `stripe.duck.js`. If you want to keep using the custom form in your own application, you can take these deprecated components as a starting point but you should keep in mind that **they will not be updated** by our team. This means you need to take care that you collect all the required information and somehow take care that it's possible to fill in information if that's needed later.

You can find the deprecated files still from v.3.7.0

- [`PayoutDetailsForm`](https://github.com/sharetribe/ftw-daily/tree/v3.7.0/src/forms/PayoutDetailsForm)
- [`PayoutPreferencesPage`](https://github.com/sharetribe/ftw-daily/tree/v3.7.0/src/containers/PayoutPreferencesPage)
- [`stripe.duck.js`](https://github.com/sharetribe/ftw-daily/blob/v3.7.0/src/ducks/stripe.duck.js).

### Creating or updating the Stripe account

=> `sdk.stripeAccount.create`

Create a new Stripe account with `country` and `bankAccountToken`.

- Note: we don't need to pass accountToken to Flex API anymore because all the account information is handled in Stripe onboarding

**OR**

=> `sdk.stripeAccount.update`

Update the Stripe account with `bankAccountToken`.

- Note: currently Stripe doesn't support updating the country

### Fetching the information about Stripe account

=> `sdk.stripeAccount.fetch`

If the user already has the Stripe account you need to fetch the up-to-date account data from the API. This same data is returned after create and update Stripe account API calls.

### Creating Account Link

=> `sdk.stripeAccountLink.create`

Depending on the status of the Stripe account we will show `StripeConnectAccountStatusBox` with "Get verified" or "Edit Stripe account link". When the user clicks the link, a new Account Link is created and the user is redirected to Stripe

When the user returns to the application we need to check the URL parameter of the return URL.

- If the user returned to the success URL, we need to check the status of the Stripe account. Returning to success URL doesn't mean that verifying the Stripe account was successful. It only means that nothing unexpected happened.
- If there is no requirement missing from the Stripe account show the status box in `verificationSuccess` mode
- If there is still something missing from the Stripe account show the status box in `verificationNeeded` mode
- If the user returned to failure URL, show the status box in `verificationError` mode
  - failure happens e.g. if the page was refreshed or if the link was already expired


OtterleyW:
Accidentally closed the branch when I rebase the master and there were no new commits yet. The branch should be open again when we add new commits here.

PR #1235, opened 2019-11-28
Deprecate PayoutPreferencesPage and PayoutDetailsForm

OtterleyW:
Deprecate `PayoutPreferencesPage` and `PayoutDetailsForm`. These files will not be updated after we move to Stripe's [Connect Onboarding](https://stripe.com/docs/connect/connect-onboarding) flow.  New `StripePayoutPage` will be used instead of `PayoutPreferencesPage`.

Also, add a new list of configs to `stripe-config.js` since we are not going to need all the specified information in the old one. This means that import in `config.js` is changed too.


src/components/StripeBankAccountTokenInputField/StripeBankAccountRequiredInput.js
```diff
@@ -43,7 +44,7 @@ const StripeBankAccountRequiredInput = props => {
   const errorMessage = <p className={css.error}>{inputError}</p>;
 
   return (
-    <div className={classes}>
+    <div className={wrapperClassName ? wrapperClassName : classes}>
```
Gnito: I think the existing pattern of `rootClassName` is the correct way to deal with this.

If success/error classes should not be used, then those classes could be used through prop that disables them (e.g. `useValidationClasses=<true/false>`)

src/components/StripeBankAccountTokenInputField/StripeBankAccountTokenInputField.css
```diff
@@ -5,17 +5,32 @@
 }
 
 .root {
+  display: flex;
+  justify-content: space-between;
+  flex-shrink: 0;
```
Gnito: push this to the first styling rule.
```
.root {
  /* Parent component should not shrink this field */
  flex-shrink: 0;

  /* This component uses flexbox layout too */
  display: flex;
  justify-content: space-between;
  flex-wrap: wrap;
}
```

src/components/StripeBankAccountTokenInputField/StripeBankAccountTokenInputField.js
```diff
@@ -260,11 +260,16 @@ class TokenInputFieldComponent extends Component {
 
     const inputConfiguration = requiredInputs(country);
 
+    // E.g. Japan has 6 fields in the bank account details so we want to
+    // show the inputs in two columns on bigger screens
+    const longForm = inputConfiguration.length >= 6;
```
Gnito: Good thinking 👍 
(Could you create a constant: `const MIN_INPUT_COUNT_FOR_TWO_COLUMNS = 6;` - i.e. all the magic numbers to the beginning of the file.)

src/ducks/stripe.duck.js
```diff
@@ -273,6 +273,154 @@ export const retrievePaymentIntentError = payload => ({
...
+    const stripeMessage =
+      e.apiErrors && e.apiErrors.length > 0 && e.apiErrors[0].meta
+        ? e.apiErrors[0].meta.stripeMessage
+        : null;
+    log.error(err, 'create-stripe-account-failed', { stripeMessage });
+    throw e;
+  });
+};
+
+// ================ Deprecated thunk functions ================ //
```
Gnito: I wonder if we should comment these functions out - that way they are not in production build.
Another option would be to create some extra file: `stripeInAppOnboarding.duck.js`, that uses these functions and has alert on top of the file stating that we are not going to keep this Stripe API onboarding approach up-to-date - but focus on hosted onboarding.

src/forms/PayoutDetailsForm/PayoutDetailsForm.js
```diff
@@ -1,3 +1,15 @@
+/** NOTE: This code will not be updated!
+ *
+ * If you want to use this form you need to add the import back to `forms/index.js`
+ *
+ * We are currently using Connect Onboarding flow provided by Stripe.
```
Gnito:
```suggestion
 * We are currently using hosted Connect Onboarding flow provided by Stripe.
```

src/stripe-config.js
```diff
@@ -1,19 +1,299 @@
...
+
+Remember to change the import from config.js if you want to use this config! 
+
+This setup is for API version '2019-02-19' and later.
+If you have an older API version in use, you need to update your Stripe API.
+You can check your Stripe API version from Stripe Dashboard -> Developers.
+Stripe only supports payments in certain countries, see full list
+at https://stripe.com/global
+*/
 export const stripeSupportedCountries = [
```
Gnito: Maybe we should comment these out too?

Or alternatively, create some kind of merge:
```js
export const stripeCountryDetails = useLegacy
  ? stripeCountryDetails.map(c => {
      const legacyConfig = stripeSupportedCountries.find(legacy => legacy.code === c.code);
      const accountConfigMaybe = legacyConfig && legacyConfig.accountConfig ? { accountConfig: legacyConfig.accountConfig } : {};
      return { ...c, ...accountConfigMaybe, /*etc.*/ };
    })
  : stripeCountryDetails;
```

src/components/EditListingWizard/EditListingWizard.js
```diff
@@ -28,8 +28,8 @@ import css from './EditListingWizard.css';
 // Show availability calendar only if environment variable availabilityEnabled is true
 const availabilityMaybe = config.enableAvailability ? [AVAILABILITY] : [];
 
-// TODO: PHOTOS panel needs to be the last one since it currently contains PayoutDetailsForm modal
-// All the other panels can be reordered.
+// You can reorder panels by reordering tabs in this array.
+// If you change the order of the panels, remember to change the texts in buttons accordinly from en.json file (keys starting with EditListingWizard.saveNew...)
```
Gnito: It's probably better to do this change separate PR. (I have a feeling that there are other issues that might need changes too - but I'm not sure...)

PR #1236, opened 2019-11-28
StripePayoutForm

OtterleyW:
Note: functionally this should work. Might need some baseline check.

localhost_3000_account_payments (3) copy
localhost_3000_account_payments (3)
localhost_3000_account_payments_
localhost_3000_account_payments_verification-failed


src/ducks/stripeConnectAccount.duck.js
```diff
@@ -0,0 +1,265 @@
...
+    .catch(err => {
+      const e = storableError(err);
+      dispatch(getAccountLinkError(e));
+      const stripeMessage =
+        e.apiErrors && e.apiErrors.length > 0 && e.apiErrors[0].meta
+          ? e.apiErrors[0].meta.stripeMessage
+          : null;
+      log.error(err, 'get-stripe-account-link-failed', { stripeMessage });
+      throw e;
+    });
```
kpuputti: This `.catch` is repeating a few times here. Would it make sense to extract (some of) it for reuse?

src/forms/StripeConnectAccountForm/StripeConnectAccountForm.js
```diff
@@ -0,0 +1,278 @@
...
+      {country ? (
+        <StripeBankAccountTokenInputField
+          className={css.bankDetailsStripeField}
+          disabled={disabled}
+          name="bankAccountToken"
+          formName="StripeConnectAccountForm"
+          country={country}
+          currency={countryCurrency(country)}
+          validate={validators.required(' ')}
+        />
```
kpuputti: Is this Field used just for visuals or is it editable? I don't understand why there's a dropdown and an input with the value of the selected option from the dropdown.

src/forms/StripeConnectAccountForm/StripeConnectAccountForm.js
```diff
@@ -0,0 +1,278 @@
...
+    </div>
+  ) : stripeAccountError ? (
+    <div className={css.error}>
+      <FormattedMessage id="StripeConnectAccountForm.createStripeAccountFailed" />
+    </div>
+  ) : null;
+};
+
+const StripeConnectAccountFormComponent = props => {
+  const [showCardUpdateInput, setShowCardUpdateInput] = useState(false);
```
kpuputti: Whee, a hook in FTW \o/

src/forms/StripeConnectAccountForm/StripeConnectAccountForm.js
```diff
@@ -0,0 +1,278 @@
...
+          ready,
+          savedCountry,
+          stripeAccountFetched,
+          stripeBankAccountLastDigits,
+          submitButtonText,
+          values,
+          stripeConnected,
+        } = fieldRenderProps;
+
+        const accountDataLoaded = stripeConnected && (stripeAccountFetched && savedCountry);
```
kpuputti: No need for the parens.

src/forms/StripeConnectAccountForm/StripeConnectAccountForm.js
```diff
@@ -0,0 +1,278 @@
...
+            ) : null}
+          </Form>
+        ) : (
+          <div className={css.missingStripeKey}>
+            <FormattedMessage id="StripeConnectAccountForm.missingStripeKey" />
+          </div>
+        );
+      }}
+    />
+  );
```
kpuputti: This whole return statement above has so much parallel and nested ternaries that it's really hard to fully understand. Could it be split down to variables above?

src/forms/StripeConnectAccountForm/StripeConnectAccountForm.js
OtterleyW: This StripeBankAccountTokenInputField is used for getting the bank account token from Stripe. It's the place where we collect the bank account number (e.g. IBAN). Because it depends on the country which fields should be shown (IBAN has only one field but e.g. Canadian bank account has three different fields) we need to ask the country first. That's why the selected option is passed to input. Also, country and currency are required information for generating the bank account token.

src/ducks/stripeConnectAccount.duck.js
OtterleyW: I decided not to do this right now but I think this is a good suggestion. I added a card to Clubhouse. It might be useful to check also stripe.duck.js in case there is also something similar so this could be extracted to a helper function.

src/containers/StripePayoutPage/StripePayoutPage.duck.js
```diff
@@ -51,18 +58,32 @@ export const savePayoutDetailsSuccess = () => ({
...
   // in case this page load fails.
-  dispatch(setInitialState());
+  dispatch(setInitialValues());
 
-  return dispatch(fetchCurrentUser());
+  return dispatch(fetchCurrentUser())
+    .then(response => {
+      const currentUser = getState().user.currentUser;
+      if (currentUser && currentUser.stripeAccount) {
+        dispatch(fetchStripeAccount());
```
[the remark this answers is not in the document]
Gnito: @kpuputti it’s not really fully thought through, but we have stripeAccountFetched flag set under the hood, when fetchStripeAccount is called. So, this gives possibility to start rendering StripePayoutPage (and EditListingWizard) before accoutData is fetched.

PR #1254, opened 2020-02-03
Login as user

lyyder:
Adds two endpoints to the server that handle authorization code grant type login requests. Enables operators to log in as a user.

Takes the new `isAnonymous` attribute returned by `sdk.authInfo` to resolve if a user is logged in.

A banner notifying a user in case their access token has limited write access will be added in a distinct PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
     "seedrandom": "^3.0.3",
-    "sharetribe-flex-sdk": "^1.8.0",
+    "sharetribe-flex-sdk": "^1.9.0",
     "sharetribe-scripts": "3.1.1",
     "smoothscroll-polyfill": "^0.4.0",
     "source-map-support": "^0.5.9",
--- a/server/apiRouter.js
+++ b/server/apiRouter.js
@@ -1,0 +1,156 @@
+/**
+ * This file contains server side endpoints that can be used to perform backend
+ * tasks that can not be handled in the browser.
+ *
+ * The endpoints should not clash with the application routes. Therefore, the
+ * enpoints are prefixed in the main server where this file is used.
+ */
+
+const http = require('http');
+const https = require('https');
+const express = require('express');
+const crypto = require('crypto');
+const sharetribeSdk = require('sharetribe-flex-sdk');
+const Decimal = require('decimal.js');
+
+const CLIENT_ID = process.env.REACT_APP_SHARETRIBE_SDK_CLIENT_ID;
+const ROOT_URL = process.env.REACT_APP_CANONICAL_ROOT_URL;
+const CONSOLE_URL =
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
     "seedrandom": "^3.0.3",
-    "sharetribe-flex-sdk": "^1.8.0",
+    "sharetribe-flex-sdk": "1.9.0",
```
Gnito: Was there a reason to remove caret?

server/index.js
```diff
@@ -132,6 +133,9 @@ if (!dev) {
   }
 }
 
+// Server-side routes that do not render the application
+app.use('/api', apiRouter);
```
Gnito: Could you also add a note to here:
https://github.com/sharetribe/ftw-daily/blob/master/src/routeConfiguration.js#L48

Just to ensure that if there's a clash of existing code in some of the customizations the first step is to look from routeConfiguration that what's wrong there. Some code comment would give a hint that people need to take a look to server/index.js too.

package.json
lyyder: No, that's just a mistake. 👍

server/apiRouter.js
```diff
@@ -0,0 +1,135 @@
...
+response_type=code&\
+client_id=${CLIENT_ID}&\
+redirect_uri=${loginAsRedirectUri}&\
+user_id=${userId}&\
+state=${state}&\
+code_challenge=${codeChallenge}&\
+code_challenge_method=S256`;
+
+  res.cookie(stateKey, state);
+  res.cookie(codeVerifierKey, codeVerifier);
```
Gnito: I assume that SDK clears these tokens automatically when logout happens. I didn't test though.

server/apiRouter.js
```diff
@@ -0,0 +1,135 @@
...
+ * replacing unaccepted characters.
+ */
+const urlifyBase64 = base64Str =>
+  base64Str
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=/g, '');
+
+// Creates a 302 response that redirects to start an OAuth2 Authorization code
+// login flow.
```
Gnito: Could you elaborate more here? I.e. what are the steps/redirections in this process.

lyyder: I'll try.

server/apiRouter.js
bladealslayer: Also, these two cookies should be flagged as secure, if possible (e.g. when HTTPS is enabled). Shouldn't they also have some (short) expiration time? Were there nowadays any other recommended security-related flags or settings for cookies?

lyyder: Added a 30 second ttl to the cookies and retract them after they've been read.

PR #1259, opened 2020-02-14
Show a banner when user is logged in with limited access

lyyder:
Show a banner when a user is logged in with limited access. In practice this means that an operator is logged in on behalf of a marketplace user. In such case the access token received from Auth API has scope `user:limited`

The banner is tied to the Topbar component. It is hidden on SearchPage as there it would push the top bar on top of the search panel.

Desktop:

Screenshot 2020-02-14 at 11 45 43

Mobile:

Screenshot 2020-02-14 at 11 45 55

### DIFF
--- a/src/components/LimitedAccessBanner/LimitedAccessBanner.css
+++ b/src/components/LimitedAccessBanner/LimitedAccessBanner.css
@@ -1,0 +1,31 @@
+@import '../../marketplace.css';
+
+.root {
+  background-color: #df492a;
+  text-align: center;
+  padding: 10px 20px 9px;
+}
+
+.text {
+  margin: 0;
+  display: inline-block;
+  color: #fff;
+  font-size: 16px;
+  margin-bottom: 16px;
+  line-height: 20px;
+}
+
+.button {
+  background: #2a3d4b;
+  margin: 0 16px;
+  padding: 8px 16px;
+  border-radius: 4px;
+  font-size: 14px;
+  color: #fff;
+  border: 0;
+
+  &:hover {
+    text-decoration: none;
+    background: #364f61;
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2020-XX-XX
 
-- [add] Support for logging in as a user from Console. [#1254](https://github.com/sharetribe/ftw-daily/pull/1254)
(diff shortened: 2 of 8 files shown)


src/translations/de.json
```diff
@@ -318,6 +318,8 @@
   "LayoutWrapperAccountSettingsSideNav.passwordTabTitle": "Passwort",
   "LayoutWrapperAccountSettingsSideNav.paymentsTabTitle": "Zahlungseinstellungen",
   "LayoutWrapperAccountSettingsSideNav.paymentMethodsTabTitle": "Zahlungseinstellungen",
+  "LimitedAccessBanner.message": "Your are logged in as {firstName} {lastName}. You have limited rights for actions.",
+  "LimitedAccessBanner.logout": "Log out",
```
OtterleyW: No need to add the English texts to other translation files. In case some key is missing from the translation file we will default to the English version in the code. (Same with es.json and dr.json)

src/components/LimitedAccessBanner/LimitedAccessBanner.css
```diff
@@ -0,0 +1,31 @@
...
+.root {
+  background-color: #df492a;
+  text-align: center;
+  padding: 10px 20px 9px;
+}
+
+.text {
+  margin: 0;
+  display: inline-block;
+  color: #fff;
```
OtterleyW: You could use `--matterColorLight` here (it's also #ffffff)

src/components/LimitedAccessBanner/LimitedAccessBanner.css
```diff
@@ -0,0 +1,31 @@
+@import '../../marketplace.css';
+
+.root {
+  background-color: #df492a;
```
OtterleyW: I'm not sure if this color should come from `marketplace.css` so that it would be easier to change to be in line with other marketplace colors. On the other hand, I guess the notification should only be visible to the operator so not sure if this is really needed.

lyyder: Good reminder regarding the colors. I think it makes sense to use the `--marketplaceColor` in the background. The grey button however, is a bit hard to make work with the colors from marketplace.css I think with it we can stick to the custom colors.

lyyder: On second thoughts, if the banner bg color is the marketplace color it might go unnoticed. Therefore, it's good that it's a notable color and differs from the marketplace main color. I'll keep the custom colors as this is meant to communicate a special behaviour.

PR #1265, opened 2020-02-28
Phase - Create faq page

tbsphpdev:
Create faq page with accordion

Resolves: #
See also: #


Gnito:
@tbsphpdev, I assume that this is an accidental pull request to upstream repo.

PR #1269, opened 2020-03-12
Redirect user back to Stripe on returning to failure url

sktoiva:
When creating account links and sending user to Stripe, there's a possibility that user is sent back to FTW to failure url. This can happen if the browser is refreshed (e.g. during language change in the Connect Onboarding UI), link has been used, expired or the account has been rejected. Stripe recommends that in these cases, a new account link is generated and user redirected back to Stripe. If there's something bigger failing, the account link generation will fail and that stops the redirection back to Stripe. More info here: https://stripe.com/docs/connect/connect-onboarding#failureurl

This pull request implements the above behavior in order to create seamless experience for the user.

image

image

image

image


src/components/EditListingWizard/EditListingWizard.js
```diff
@@ -27,6 +27,7 @@ import EditListingWizardTab, {
   PHOTOS,
 } from './EditListingWizardTab';
 import css from './EditListingWizard.css';
+import { useEffect } from 'react';
```
OtterleyW: This could actually be combined with the other import from `react` in the first line

```suggestion
import React, { Component, useEffect } from 'react';
```

src/components/EditListingWizard/EditListingWizard.js
```diff
@@ -263,6 +272,8 @@ class EditListingWizard extends Component {
       fetchStripeAccountError,
       stripeAccountFetched,
       stripeAccount,
+      stripeAccountError,
+      stripeAccountLinkError,
```
OtterleyW: When adding new props to component we should also make sure that these new props are added to propType check in the end of this file. I have really bad habit of forgotting these myself so propably I didn't mention it yesterday.

PropTypes are checked here: https://github.com/sharetribe/ftw-daily/blob/b5ad36f5427f1d664e021ce86b3118bb69a3c9a0/src/components/EditListingWizard/EditListingWizard.js#L476
These new props should not be required so we can also add default value (null) for them: 
https://github.com/sharetribe/ftw-daily/blob/b5ad36f5427f1d664e021ce86b3118bb69a3c9a0/src/components/EditListingWizard/EditListingWizard.js#L469

We should check the propTypes also on `EditListingPage` and `PayoutDetailsPage`

src/components/EditListingWizard/EditListingWizard.js
```diff
@@ -384,54 +405,59 @@ class EditListingWizard extends Component {
...
+                  })}
+                  stripeAccountError={stripeAccountError}
+                  stripeAccountFetched={stripeAccountFetched}
+                  stripeAccountLinkError={stripeAccountLinkError}
+                  onChange={onPayoutDetailsFormChange}
+                  onSubmit={rest.onPayoutDetailsSubmit}
+                  onGetStripeConnectAccountLink={handleGetStripeConnectAccountLink}
+                  stripeConnected={stripeConnected}
+                >
+                  {stripeConnected && (!returnedAbnormallyFromStripe && showVerificationNeeded) ? (
```
OtterleyW: We don't need the parenthesis here

```suggestion
                  {stripeConnected && !returnedAbnormallyFromStripe && showVerificationNeeded ? (
```

PR #1277, opened 2020-03-23
Sort search results

kpuputti:
This PR adds search results sorting to the SearchPage. Results can be sorted with these criteria:

- Newest
- Oldest
- Lowest price
- Highest price
- Relevance (Keyword search)

If there is a keyword filter, the relevance option is automatically selected and the other options are disabled. If there is no keyword, the relevance filter is disabled.

**NOTE:** The results count was moved to the left side of the sorting dropdown after these screenshots were taken.

## Default view

Screenshot 2020-03-23 at 16 01 20

## Mobile view

Screenshot 2020-03-23 at 16 00 09

## Mobile view with keyword filter

Screenshot 2020-03-23 at 16 00 19

## Desktop view

Screenshot 2020-03-23 at 16 00 47

## Desktop view with keyword filter

Screenshot 2020-03-23 at 16 01 04

## Customization

The sorting is turned on by default. It can be disabled in the `marketplace-custom-config.js` with the `sortConfig.active` option.

### DIFF
--- a/src/components/SortBy/SortByPopup.css
+++ b/src/components/SortBy/SortByPopup.css
@@ -1,13 +1,6 @@
 @import '../../marketplace.css';
 
 .root {
-  display: inline-block;
-  margin-left: auto;
-  margin-right: 8px;
-
-  &:last-of-type {
-    padding-right: 0;
-  }
 }
 
 .icon {
@@ -84,7 +77,6 @@
   color: var(--matterColor);
   margin-left: 30px;
   margin-right: 30px;
-  margin-bottom: 24px;
 }
 
 .menuItem {
--- a/src/components/SortBy/SortByPopup.css
+++ b/src/components/SortBy/SortByPopup.css
@@ -1,0 +1,150 @@
+@import '../../marketplace.css';
+
+.root {
+  display: inline-block;
+  margin-left: auto;
+  margin-right: 8px;
+
+  &:last-of-type {
+    padding-right: 0;
+  }
+}
+
+.icon {
+  position: relative;
+  font-size: 10px;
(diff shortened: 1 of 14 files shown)


src/marketplace-custom-config.js
```diff
@@ -64,3 +64,24 @@ export const dateRangeFilterConfig = {
 export const keywordFilterConfig = {
   active: true,
 };
+
+const relevanceKey = 'relevance';
```
Gnito: Add some comment here too. Since this a config file, it might even be better to keep the `relevanceKey: 'relevance'` inside the sortConfig variable.

src/components/SortBy/SortByPopup.js
```diff
@@ -0,0 +1,120 @@
+import React, { Component } from 'react';
+import { string, func, arrayOf, shape, number } from 'prop-types';
+import classNames from 'classnames';
+
+import { Menu, MenuContent, MenuItem, MenuLabel } from '..';
```
Gnito:
```suggestion
import { Menu, MenuContent, MenuItem, MenuLabel } from '../../components';
```

src/components/SortBy/SortByPopup.css
```diff
@@ -0,0 +1,150 @@
...
+  position: absolute;
+  top: 2px;
+  left: 0px;
+  height: calc(100% - 7px);
+  width: 6px;
+  background-color: var(--matterColorDark);
+}
+
+.menuHeading {
+  /* @apply --marketplaceListingAttributeFontStyles; */
```
Gnito: Should this be removed?

PR #1279, opened 2020-03-27
Fix "Cannot POST" error for PaymentMethodsForm:

rush1506:
- When using npm, the null handleSubmit would override the FinalForm handleSubmit method
- This would cause error when the form is submit  because the formRenderProps give out null handleSubmit function


Gnito:
Thanks, and sorry - this PR has been here unnoticed for a while. I'll check this code soonish.

Gnito:
I close this one since the actual commit was merged in different PR.

PR #1280, opened 2020-03-27
Improve search filter UI

kpuputti:
This PR improves the search page sorting controls for different screen sizes.

## Mobile

Screenshot 2020-03-30 at 13 31 54

## Desktop

Screenshot 2020-03-30 at 13 32 14

### DIFF
--- a/src/components/SearchFilters/SearchFilters.css
+++ b/src/components/SearchFilters/SearchFilters.css
@@ -18,7 +18,7 @@
   }
 }
 
-.resultOrganization {
+.searchOptions {
   display: flex;
   flex-direction: row;
   margin-bottom: 24px;
--- a/src/components/SearchFilters/SearchFilters.css
+++ b/src/components/SearchFilters/SearchFilters.css
@@ -4,10 +4,6 @@
   display: flex;
   flex-direction: column;
   background-color: var(--matterColorBright);
-
-  @media (--viewportLarge) {
-    flex-direction: row;
-  }
 }
 
 .longInfo {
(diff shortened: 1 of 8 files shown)


src/components/SearchFilters/SearchFilters.css
```diff
@@ -24,17 +18,19 @@
   }
 }
 
+.resultOrganization {
```
Gnito: Sounds company level stuff and about results instead of filters and sorting. 
How about: 'searchCustomization', 'queryParameters', or 'searchOptions'?

src/components/MenuContent/MenuContent.css
```diff
@@ -1,14 +1,20 @@
...
   z-index: var(--zIndexPopup);
 
   background-color: var(--matterColorLight);
   border-top: 1px solid var(--matterColorNegative);
   box-shadow: var(--boxShadowPopup);
   border-radius: 0 0 2px 2px;
   transition: var(--transitionStyleButton);
+
+  /* On mobile, the popup will take full screen width. */
+  position: fixed;
```
Gnito: Does this mean that the filter will move along when user is scrolling and the parent component has enough height for some reason?

How about using `width: 100vw;` instead?

Gnito: or calc(100vw - 48px)

kpuputti: Hmm, the top seems to stay where the parent was when the dropdown is rendered. Even if the page is scrolled and the parent moves. In search page this is ~ok since the parent becomes sticky and doesn't move that much.

The full width itself isn't enough, the coordinates for absolute position is in relation to the parent, not the screen. So calculating the horizontal position from the parent is a lot more complex.

kpuputti: This is now fixed, see the latest commits.

PR #1287, opened 2020-05-12
Remove string cleaning from state-saved values.

Gnito:
This StripeBankAccountTokenInputField component is very old code (in a sense that there has been a lot of Stripe iteration and addition of new countries) - we should probably rewrite this component at some point. It's not even following our pretty old convention of naming Final Form field components as Field*.

This PR removes:
- String cleaning for state-saved values. 
  It is done only on code values (bank codes, routing numbers, account-number, etc.) sent to Stripe. 

  I wonder if that formatted code was originally used for some error printing (i.e. sent values). 

- Removed `errorsClearedFromInputs` state handling after the async call was made.
  This was dead code and just caused saving of duplicate info to state.

### DIFF
--- a/src/components/StripeBankAccountTokenInputField/StripeBankAccountTokenInputField.js
+++ b/src/components/StripeBankAccountTokenInputField/StripeBankAccountTokenInputField.js
@@ -9,7 +9,6 @@ import config from '../../config';
 
 import {
   BANK_ACCOUNT_INPUTS,
-  cleanedString,
   formatFieldMessage,
   requiredInputs,
   mapInputsToStripeAccountKeys,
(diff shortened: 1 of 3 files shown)


src/components/StripeBankAccountTokenInputField/StripeBankAccountTokenInputField.js
```diff
@@ -150,19 +149,16 @@ class TokenInputFieldComponent extends Component {
...
         );
         const valuesAreUnchanged = changedValues.length === 0;
 
         // Handle response only if the input values haven't changed
         if (this._isMounted && valuesAreUnchanged) {
           this.setState(prevState => {
-            const errorsClearedFromInputs = inputsNeeded.map(inputType => {
-              const input = prevState[inputType];
-              return { ...input, error: null };
-            });
```
Gnito, line 164: This code actually creates another set of input values to state - it's about array vs object: `state[0]` instead of overwriting `state[inputType]`. Reducer would easily solve this, but in the end, this is dead code, since error clearing is done earlier.

OtterleyW:
LGTM! Remember to update the changelog

PR #1293, opened 2020-05-18
Ehsan

farrox:


farrox:
aa

OtterleyW:
@farrox this seems to be accidental PR from your customization project so I'm closing this.

Also, I would recommend that you generate new API keys for Flex, Mapbox, and Stripe because you pushed the .env file here! You should not add that file to your GitHub repo.

PR #1295, opened 2020-05-19
wip: util functions for constructing lineItems based on pricing config

OtterleyW:


src/util/lineItems.js
```diff
@@ -0,0 +1,168 @@
...
+  const providerLineItems = lineItems.filter(lineItem => lineItem.includeFor.includes('provider'));
+  return calculateTotalFromLineItems(providerLineItems);
+};
+
+export const calculateTotalForCustomer = lineItems => {
+  const providerLineItems = lineItems.filter(lineItem => lineItem.includeFor.includes('customer'));
+  return calculateTotalFromLineItems(providerLineItems);
+};
+
+export const constructLineItems = (startDate, endDate, unitPrice) => {
```
kpuputti, line 125: Since customizations likely add new booking params, it would make sense to pass around a generic `bookingData` object that is used to construct the line items for the estimation and for the actual tx initiation in the backend. This way adding a new field to the booking form doesn't require changing all the places that pass around the booking data.

src/util/lineItems.js
```diff
@@ -0,0 +1,168 @@
...
+  const bookingLineItem = validLineItem({
```
kpuputti, line 126: Instead of wrapping each line item with the `validLineItem` function, you could collect all line items into an array, and then iterate each item with the function:

```js
const lineItem1 = { /* ... */ };
const lineItem2 = { /* ... */ };
const lineItem3 = { /* ... */ };

const lineItems = [
  lineItem1,
  lineItem2,
  lineItem3,
];

return lineItems.map(validLineItem);
```

You can even skip adding each item into a variable, and just construct them inline within the `lineItems` array.

src/util/lineItems.js
```diff
@@ -0,0 +1,168 @@
...
+    code: 'line-item/cleaning-fee',
+    unitPrice: new Money(7500, 'USD'),
+    quantity: 1,
+    includeFor: ['customer', 'provider'],
+  });
+
+  const providerCommission = validLineItem({
+    code: 'line-item/provider-commission',
+    unitPrice: calculateTotalFromLineItems([bookingLineItem, cleaningFee]),
+    percentage: -15,
```
kpuputti, line 143: These commission percentages could be defined as variables within this file. That way they are easily customized without looking at the line item generation, and we can also give good documentation on the possible values.

src/util/lineItems.js
kpuputti, line 125: Idea: could this file only have a single function: `transactionLineItems(bookingData)`? That way it's quite simple what part should be customized, and the helper functions would be in another file.

src/util/lineItems.js
OtterleyW: The reason why they are now wrapped separately is that some lineItems (e.g. commissions) need to know the lineTotal of other lineItems and that's calculated inside validLineItem function. 

Maybe a better solution would be to add new helper function for calculating the lineTotal and have other validations separately for all of the lineItems like in your example.

src/util/lineItems.js
OtterleyW: I think this is a good idea!

src/util/lineItems.js
kpuputti: Ah ok. Validation separate from line item creation would be nice, but also not a huge problem in the current solution.

OtterleyW:
Closed in favor of #1299 and #1305

PR #1296, opened 2020-05-20
Refactoring filter setup

Gnito:
Current filter setup is a mess since routing and state handling (for user inputs) is spread out to 3 different container components: 
- SearchFilters (primary filters on desktop)
- SearchFiltersPanel (secondary filters on desktop)
- SearchFiltersMobile

Furthermore, creating different versions of generic enum-type filters is difficult since you need to duplicate code on those different containers. (e.g. amenities filter is using SelectMultipleFilter, but it could be configurable instead of fixed to variable names etc.)

**marketplace-custom-config.js** contains filter setup (and sort by too).

```js
export const filters = [
  {
    id: 'price',
    label: 'Price',
    type: 'PriceFilter',
    group: 'primary',
    queryParamNames: ['price'],
    config: {
      min: 0,
      max: 1000,
      step: 5,
    },
  },
  {
    id: 'category',
    label: 'Category',
    type: 'SelectSingleFilter',
    group: 'secondary',
    queryParamNames: ['pub_category'],
    config: {
      options: [
        { key: 'smoke', label: 'Smoke' },
        { key: 'electric', label: 'Electric' },
        { key: 'wood', label: 'Wood' },
        { key: 'other', label: 'Other' },
      ],
    },
  },
  {
    id: 'view',
    label: 'View',
    type: 'SelectSingleFilter',
    group: 'secondary',
    queryParamNames: ['pub_view'],
    config: {
      options: [
        { key: 'sea', label: 'Sea' },
        { key: 'lake', label: 'Lake' },
        { key: 'forest', label: 'Forest' },
        { key: 'garden', label: 'Garden' },
        { key: 'other', label: 'Other' },
      ],
    },
  },
//etc.
```
So, every filter are described in an array (to get rendering order) and they have keys
- **id**: id of the filter
- **label**: faster customization without en.json setup
- **type**: represents one of the existing filter components
- **group**:  is this 'primary' or 'secondary' filter
- **queryParamNames**: describes parameters to be used with API (e.g. _price_ or _pub_amenities_)
- **config** extra configuration that the filter component needs.

In addition, sort configuration also exists in the `marketplace-custom-config.js`.

----

Then all the filter components are added through a single file (`SearchPage/FilterComponent.js`) and related code is removed from those child-containers (i.e. SearchFilters, SearchFiltersPanel, SearchFiltersMobile). The idea is that when you add a custom filter component (let's say, _LongFilter_ since we don't have filter component for long type), you only need to touch this one file instead of spreading and duplicating code in different UI containers.

In practice, this means that those container components just render filters through `props.children` and their main tasks are to provide UI and _Apply_, _Reset all_, etc. buttons.

In the SearchPage/MainPanel.js, those presentational containers get their children rendered through one generic component (called `FilterComponent`), which essentially just maps correct filter config, state/submit handling functions, and other props that the UI container expects.

```jsx
        <SearchFiltersPrimary {/* what ever props the container needs*/}>
          {filterConfig.filter(selectPrimaryFilters).map(config => {
            return (
              <FilterComponent
                key={`SearchFiltersPrimary.${config.id}`}
                idPrefix="SearchFilters"
                filterConfig={config}
                urlQueryParams={urlQueryParams}
                initialValues={this.initialValues}
                getHandleChangedValueFn={this.getHandleChangedValueFn}
                showAsPopup
                contentPlacementOffset={FILTER_DROPDOWN_OFFSET}
              />
            );
          })}
        </SearchFiltersPrimary>
```

----

Renaming some of the existing components makes sense at this point: 
- `SearchFilters` -> `SearchFiltersPrimary`
- `SearchFiltersPanel` -> `SearchFiltersSecondary`.

----

SortBy's state is now also tracked in MainPanel - so it's not cleared when filters are updated.

----

Taking update from upstream:
if you have just modified filters on SearchPage, it might be best if you reimplement those filter configurations through marketplace-custom-config.js - i.e. skip the merge conflicts since customizing existing filters (add/remove/update/reorder) is very easy with this new setup.

If you have create completely new filters, you should note that in this new setup initialValues for the filter are strings that come from URL parameters. Filter component needs to parse them themselves and vice versa: onSubmit should give formatted URL parameter values per key (e.g. `{ price: '15,1000' }`).


ovan:
@Gnito This looks really good! A couple of comments / questions / suggestions:

1.  **label** - Can I still use translation mechanism to pull in the value for the label if I want to? I guess nothing stopping that since this is a .js file and not a .json? Is that just one require + function call away?
2. **paramName** - I propose renaming this to **queryParam** or **queryParamName** to make it clear how this is used and what it affects.
3. I like the array format to set order. It's very clear and requires no duplication / modifying multiple places.
4. I'm interested to see the handling of has_any / has_all. That's part of the API and spec now so we should support that, and also by default always explicitly set it when building query param values for enum filters instead of relying on server defaults.
5. Is sort defined through the same mechanism, just another filter that sets the sort query params? At least it works a little differently in the sense that you can have multiple sorts but you don't configure the query param to use. Instead you end up compose like: `sort=createdAt,-pub_gears,price` https://www.sharetribe.com/api-reference/marketplace.html#sorting
6. What about cross filter dependencies? If I use keyword search I want to disable filter control. How would that work in this? For example, `<SearchFilters>` could take a function that gets called with current composed selections and could, by **id** (which we'd need to add to filter configuration) disable a filter.

kpuputti:
IMO this looks good with the features it currently has. I also think that the array to define the order is definitely more convenient than any other hierarchical solution.

Some questions:

1. Is it possible to define filters that have multiple form components? Think about a day picker where the month and day are separate dropdowns. Or is this just an internal detail of the filter component?
2. Is it possible that a single filter controls multiple query params?
3. Can the filter just read the current values of the other filters, i.e. depend on other filters and change its behaviour based on those?
4. Is the query param name supposed to be fully configurable? Currently it kinda is, but not fully. I remember that originally we wanted to support localising the param names, but I don't think that is a common customization. It's a lot clearer to map then directly to the Marketplace API param names.

   I'm mainly asking this since if filters depend on other filter (like the sorting depending on the keyword filter), they need to read the param from the current filter values using the configuration. This is where we currently break away from having fully configurable param names as some places still have the names hard coded.

Gnito:
@ovan 

1. Yes. It should work like that using `<FormattedMessage id="some.translation.key.here" />`. Current codebase has hard-coded labels in a filter component, but the labels for options are coming from config. This would make the translations of single filter more consistent.

2. Sounds good. (How about even more clear: queryParamNameForAPI?)

4. There are an easy way and a more difficult one:
  - Before making the query to API, we just add the correct search type to that particular query parameter
  - If we try to use those search types in the filter / client app's URL, we might need to refactor the inner guts of those filters quite much since they use checkboxes' default behaviour to construct URLSearchParameter's value. So, this is potentially a bigger task (I'm not sure if there's any happy surprises that would make this easier.)

5. I was considering leaving the sortConfig in a [separate exported variable](https://github.com/sharetribe/ftw-daily/blob/master/src/marketplace-custom-config.js#L68).

6. If there's any information shared between filters, it would happen in MainPanel. (Those child containers don't care about their own children aka filters in this setup.) 

I'm a bit worried about how big a task this is going to be.
- If filter components are responsible to react to this new info, we need to refactor every component. Check for example [SortBy component / isKeywordFilterActive](https://github.com/sharetribe/ftw-daily/blob/master/src/components/SortBy/SortBy.js#L11). 
- We could just show/hide filter if some active filter's `config.hideFilters: [id1, id2]` states so - but that would also mean that the query params of those disabled filters need to be cleared. 

```js
const primaryFilters = filterConfig
  .filter(selectVisibleFilters)
  .filter(selectPrimaryFilters)
  .map(f => <FilterComponent />);
```
I'm a bit worried about the possible complexity of ´selectVisibleFilters`. To me, it sounds that this kind of "hide these filters" cross-referencing could become cumbersome to deal with (users could create loops there) - which becomes annoying if they start to directly manipulate shared URLs.

Gnito:
@kpuputti 

1. to my understanding, it would be more or less internal issue for that filter (also initialValues / onSubmit needs to be customized accordingly) 

2. Good question. I wasn't planning to change your BookingDateRangeFilter, so it should work pretty much similarly as it does with the current setup.

3. This touches the question Olli asked (Q6). To me, it sounds a bad approach to couple filters directly with each others - but customizers have full power to code what they want.

4. I'm not sure we have hard-coded param names other than this:
https://github.com/sharetribe/ftw-daily/blob/master/src/containers/SearchPage/SearchPage.js#L54

My plan was to get rid of it. So, if we start to reference filters, then we need to add ids to those individual filter configs and find relevant queryParameterName through the config file as discussed in the Olli's Q6. What comes to SortBy config, it would require something like `config.disableIfFiltersActive: [id1, id2]` => find relevant queryParamNames => convert to props.

ovan:
@Gnito 

> 5. I was considering leaving the sortConfig in a separate exported variable.

Hmmm. I think the line between filters and sort is muddy so wouldn't it make sense to have them be part of the same system. Like keyword filter is not only a filter but a sort too. Maybe that's not good enough reason though but in any case there are dependencies between the two and that has to be handled somehow.

> 3. This touches the question Olli asked (Q6). To me, it sounds a bad approach to couple filters directly with each others - but customizers have full power to code what they want.

I too was not thinking that this should be a per filter thing. Like a filter should not need to know about any other filters. I was thinking this would a place above them. A function that gets called with the combined state of all filters (not with a query string but a map keyd by filter id?) before props are passed to filter rendering. A filter would then just take a prop in it's config `isDisabled?` or similar and all it would have to know how to do is to render itself as disabled or active.

> There are an easy way and a more difficult one:
> Before making the query to API, we just add the correct search type to that particular query parameter
> If we try to use those search types in the filter / client app's URL, we might need to refactor the inner guts of those filters quite much since they use checkboxes' default behaviour to construct URLSearchParameter's value. So, this is potentially a bigger task (I'm not sure if there's any happy surprises that would make this easier.)

I don't think I'm fully following now. The `has_all` `has_any` is part of the value the filter pushes up so feels sensible to be able to prepend that to the filter's output value on filter level, right?

This makes me think, what layer is now responsible of taking all the filter output and creating a map that's in passed to the dispatched redux action? I think I just need to see more code on how this is all put together to understand this part so we can also continue this discussion once we're there.

Gnito:
@ovan If the sort config is among filter configs, then we need to extract it away in all the places where we want to render filters (because SortBy is not rendered in the same place in the UI). This is pretty consistent between any other marketplace and e-commerce sites too - different functionality, so the gestalt theory dictates that it should be separated.

Gnito:
Status update:

The key files in this new setup are:
- **marketplace-custom-config.js** (Current [dev version](https://github.com/sharetribe/ftw-daily/blob/eb6fd6646ecad5eb56af483863529f5aaa8e9e6d/src/marketplace-custom-config.js))
- SearchPage/**FilterComponent.js** (Current [dev version](https://github.com/sharetribe/ftw-daily/blob/eb6fd6646ecad5eb56af483863529f5aaa8e9e6d/src/containers/SearchPage/FilterComponent.js))

Through marketplace-custom-config.js one can add/remove/order/configure existing filter components. `SelectMultipleFilter` and `SelectSingleFilter` are existing filters that work with extended data (type enum) - so, customizers can add several select filters if they have several enumerable search schemas for listing's extended data.

If there's a need to add completely new filter component (e.g. LongFilter for filtering ['long' type data](https://www.sharetribe.com/api-reference/marketplace.html#extended-data-filtering)), customizer needs to create that filter component and add it to the generic filter wrapper: **FilterComponent**. 

Every filter gets `initialValues` as URIEncoded parameter and they have the responsibility to format their updated values similarly `onSubmit`. SearchPage/**MainPanel.js** handles the state of currently modified filters. (I.e. scenario where current URL has a certain state, but a user has modified filters too without clicking _"Apply"_  to actually make new request to fetch listings).

Gnito:
In addition to previous comment, we decided to renamed UI containers:
SearchFilters => **SearchFiltersPrimary**
SearchFiltersPanel => **SearchFiltersSecondary**
(**SearchFiltersMobile** stayed the same but some wrongly named translations were updated)

I also added a bug fix to SortBy component's functionality. Now updating a filter doesn't clear sort parameter from the URL (except when keywords filter is active since it is sorting the results already with relevance).

src/marketplace-custom-config.js
```diff
@@ -1,77 +1,150 @@
 /*
  * Marketplace specific configuration.
+ *
+ * Every filter needs to have following keys:
+ * - **id**: unique id of the filter
+ * - **label**: The default label of the filter
+ * - **type**: string that represents one of the existing filter components
```
kpuputti, line 7: You could either list the possible values here, or mention that they can be seen in `FilterComponent.js`.

src/marketplace-custom-config.js
```diff
@@ -1,77 +1,150 @@
...
+ * - **group**: is this 'primary' or 'secondary' filter?
```
kpuputti, line 8: There are only two groups. Could this default to one, and be a boolean. E.g. by default filters are secondary, and the option to make them primary is `isPrimary` in the config.

It would also be really good to briefly explain here what primary and secondary mean in the UI.

src/marketplace-custom-config.js
```diff
@@ -1,77 +1,150 @@
...
+ * - **queryParamNames**: describes parameters to be used with API (e.g. 'price' or 'pub_amenities')
```
kpuputti, line 9: Is this param name used for both the FTW URL and in the Marketplace API call? Is it possible to have a localised or otherwise different param name in FTW, but keep the param name for the API?

I don't know if having them different is a useful thing to support, and people can always hack the actual param names in the duck file, but at least mentioning here that this config changes both the FTW URL and the API params would be good.

src/util/search.js
```diff
@@ -0,0 +1,38 @@
...
+ * SelectMultipleFilter needs to parse values from format
+ * "has_all:a,b,c,d" or "a,b,c,d"
+ */
+export const parseSelectFilterOptions = uriComponentValue => {
+  return uriComponentValue && uriComponentValue.indexOf(':') >= 0
+    ? uriComponentValue.split(':')[1].split(',')
+    : uriComponentValue && uriComponentValue.indexOf(':') < 0
+    ? uriComponentValue.split(',')
+    : [];
+};
```
kpuputti, line 11: I think the above function fails if one of the values contains `:`. I don't know if that's possible, but I don't think anything restricts that, right?

To me the nested ternaries are really hard to follow. I get that avoiding naming things is good, but this is pushing way too much logic into a single expression IMHO.

src/util/search.js
```diff
@@ -0,0 +1,38 @@
...
+
+const flatten = (acc, val) => acc.concat(val);
```
kpuputti, line 13: I think `append` would be a more correct name for this function. It's not flattening a multi-dimensional array into a one-dimensional array as the name implies.

src/util/search.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Check if the filter is currently active.
+ */
+export const findActiveFilter = (filterIds, urlQueryParams, filterConfigs) => {
+  const activeQueryParamKeys = filterConfigs
+    .filter(config => filterIds.includes(config.id))
+    .map(config => config.queryParamNames)
+    .reduce(flatten, []);
```
kpuputti, line 22: A matter of taste, but as JS array ops are not efficiently collected into a single pass, you could achieve all this with a single `reduce` instead of going through the array three times.

src/util/search.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const paramEntries = Object.entries(urlQueryParams);
+  const activeKey = paramEntries.find(entry => {
+    const [key, value] = entry;
+    return activeQueryParamKeys.includes(key) && value != null;
+  });
+  return activeKey;
+};
```
kpuputti, line 30: I'm quite confused of this function. What does it do? The comment says:

> Check if the filter is currently active.

but you don't pass it a single filter to check if it's active.

Then, it only returns the first active filter it finds. Is that supposed to work like that?

src/containers/SearchPage/README.md
```diff
@@ -60,7 +64,8 @@ if SearchMap's bounds have changed enough.
 
 ### Search filters
 
-See the [filters documentation](../../../docs/search-filters.md).
+See the
+[filters documentation](https://www.sharetribe.com/docs/cookbook-search/change-search-filters-in-ftw/).
```
kpuputti, line 68: I assume this article will be updated before releasing this?

src/containers/SearchPage/README.md
```diff
@@ -36,20 +36,24 @@ possible for Topbar to take current filters into account.
...
-recommend that only 1 - 3 primary filters are passed in since they start to take too much space on
-narrow screens.
+```shell
+└── src
+    └── marketplace-custom-config.js
+```
+
+There you can set some filters to be primary filters. They are shown always on top of SearchResults
+as dropdown-selections on Desktop layout. We recommend that only 1 - 3 primary filters are passed in
+since they start to take too much space on narrow screens.
```
kpuputti, line 49: Now I see that the docs are here. Maybe refer to this file in the custom config?

src/containers/SearchPage/SearchPage.helpers.js
```diff
@@ -53,12 +61,11 @@ export const validURLParamForExtendedData = (paramName, paramValueRaw, filters)
  * @param {Object} filters Filters configuration
  */
 export const validFilterParams = (params, filters) => {
-  const filterParamNames = Object.values(filters).map(f => f.paramName);
+  const filterParamNames = filters.map(f => f.queryParamNames).reduce(flatten, []);
```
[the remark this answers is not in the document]
Gnito: @kpuputti Support of flat() wasn't there on Netlify, so I assumed that flatmap wouldn't be there either.

src/containers/SearchPage/SearchPage.helpers.js
```diff
@@ -1,44 +1,52 @@
 import intersection from 'lodash/intersection';
 import config from '../../config';
 import { createResourceLocatorString } from '../../util/routes';
+import { parseSelectFilterOptions } from '../../util/search';
 import { createSlug } from '../../util/urlHelpers';
 import routeConfiguration from '../../routeConfiguration';
 
+const flatten = (acc, val) => acc.concat(val);
```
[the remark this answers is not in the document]
Gnito: This is alternative to flat()
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

src/containers/SearchPage/MainPanel.js
```diff
@@ -1,31 +1,142 @@
...
+    };
+
+    return isArray
+      ? queryParamNames.reduce((acc, paramName) => {
+          return { ...acc, [paramName]: getInitialValue(paramName) };
+        }, {})
+      : {};
+  }
+
+  handleChangedValue(useHistoryPush) {
```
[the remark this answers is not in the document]
Gnito: @kpuputti Because it needs to have useHistory flag somewhere in the scope chain (and that flag is specific to filter instance).

Maybe it should be renamed to "getHandleChangedValue"?

src/containers/SearchPage/MainPanel.js
```diff
@@ -1,31 +1,142 @@
...
+const cleanSearchFromConflictingParams = (searchParams, sortConfig, filterConfig) => {
+  // Single out filters that should disable SortBy when an active
+  // keyword search sorts the listings according to relevance.
+  // In those cases, sort parameter should be removed.
+  const sortingFiltersActive = findActiveFilter(
+    sortConfig.conflictingFilters,
+    searchParams,
+    filterConfig
+  );
+  return sortingFiltersActive ? { ...searchParams, sort: null } : searchParams;
```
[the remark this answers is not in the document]
Gnito: `handleSortBy` seems to get the name from hard-coded value inside SortBy component. Maybe the key should be pushed to marketplace-custom-config.js

src/containers/SearchPage/MainPanel.js
```diff
@@ -38,60 +149,96 @@ class MainPanel extends Component {
...
+        mode === 'mobile'
+          ? {
+              rootClassName: css.sortBy,
+              menuLabelRootClassName: css.sortByMenuLabel,
+            }
+          : {};
+      return sortConfig.active ? (
+        <SortBy
+          {...mobileClassesMaybe}
+          sort={urlQueryParams.sort}
```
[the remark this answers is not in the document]
Gnito: previous was hard-coded too `sort={sort}` (where the variable was parsed from URL params at SearchPage.)

So, in a sense this doesn't change the current setup - but the key could be moved to marketplace-custom-config.js, which might be a better place for it.

src/marketplace-custom-config.js
Gnito: Good point about opening the concept a bit more! 

However, I wouldn't turn this to a flag. That would restrict possible customizations. We currently have 3 rendering containers for filters - it could also be more (e.g. top of results, side of results + hidden panel). It's also a good place to check the options (read about 'secondary' => type search 'secondary' from codebase).

src/marketplace-custom-config.js
Gnito: Those can be different - although, we mostly seem to follow API.
Currently, we have one exception to the rule that param names are the same in both: 'dates'. (react-dates internal output for date range is `{ dates: { startDate, endDate } }`, which is probably the reason it has been used with that filter.

I'll change the comment to this
```
 * - queryParamNames: Describes parameters to be used with queries
 *                    (e.g. 'price' or 'pub_amenities'). Most of these are
 *                    the same between webapp URLs and API query params.
 *                    You can't change 'dates', 'price', or 'keywords'
 *                    since those filters are fixed to a specific attribute.
 ```

src/util/search.js
Gnito: Yeah, reduce should be more efficient in this setup. I'm just not sure that most of our clients are familiar with `reduce`. So, lately, I have started to avoid `reduce` in places where it's more clear to have readable logic: filter > map > etc. 

Anyway, I change this to reduce format - especially, since I had to swap `flat()` call to `reduce` because Netlify/Node didn't support it.

src/util/search.js
Gnito: Hmm. Yeah, this is confusing! A bunch of code that wasn't refactored/cleand after it ended up to this 'util' file.

The original idea was to check if any of the given filters (defined by filterIds) are active. I.e. if urlQueryParams include queryParamNames defined in filterConfig. 
I'll change the function name and its variable names a bit - and I'll add a better comment.

src/containers/SearchPage/README.md
Gnito: Yes.

src/containers/SearchPage/MainPanel.js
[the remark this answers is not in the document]
Gnito: I created a new commit out of these changes, but I didn't yet merge it to previous "fix" commit as the commit name suggests.

PR #1297, opened 2020-05-20
Dev API server

kpuputti:
This PR adds a simple server that is used to start up the API routes during build time.

## TODO:

- [x] changelog


package.json
```diff
@@ -84,7 +85,8 @@
     "audit": "yarn audit --json | node scripts/audit.js",
     "clean": "rm -rf build/*",
     "config": "node scripts/config.js",
-    "dev": "node scripts/config.js --check && sharetribe-scripts start",
+    "api-server": "nodemon --watch server server/apiServer.js",
+    "dev": "node scripts/config.js --check&&concurrently -k \"sharetribe-scripts start\" \"API_SERVER_PORT=3500 yarn run api-server\"",
```
Gnito, line 89: Is it planned that you can't run the web app without api-server running on the background?

That sounds bad since it increases complexity for everybody. Could we add another script that just opens the hot-loading mode (as 'dev' did before?)

server/apiServer.js
```diff
@@ -0,0 +1,16 @@
+// Configure process.env with .env.* files
+require('./env').configureEnv();
+
+const express = require('express');
+const cookieParser = require('cookie-parser');
+const apiRouter = require('./apiRouter');
+
+const PORT = parseInt(process.env.API_SERVER_PORT, 10);
```
Gnito, line 8: I think this magic number could benefit from a comment or named variable.

package.json
Gnito: With the word 'complexity', I'm a bit worried about environments that we are not actively testing (e.g. running the app on windows)

server/apiServer.js
kpuputti: This is just the same as is done [in the main server](https://github.com/sharetribe/ftw-daily/blob/master/server/index.js#L46):

```js
const PORT = parseInt(process.env.PORT, 10);
```

Do you mean commenting on the `10` that is given to the `parseInt`? I feel that is quite usual to have it (or at least used to), and this is not really a file for customization.

package.json
kpuputti: Yep, this definitely brings in complexity. The whole point of this work is to have FTW do the price calculation, and using that with flexible pricing and privileged transitions must be done in the backend.

This does take away the possibility to only deploy statically, but IMO we haven't noticed a strong need for that, and we've also always recommended Heroku as a deployment solution. I hope this paves the way to a different setup than the CRA fork, hopefully something like Next.js in the future where backend integration is a lot simpler.

We can have a separate client-only dev server, but then the process transitions would fail, and I assume that would only increase confusion and our support that is needed. That is assuming that backend-pricing becomes the new default.

server/apiServer.js
Gnito: Yes, I was pointing to '10'. 
We had at some point an (eslint) rule that any number needs to have a variable. I think it makes sense to comment these somehow even if those strict eslint-rules have been removed in favour of Prettier.

PR #1299, opened 2020-06-01
Construct lineItems in FTW backend 

OtterleyW:
This PR adds function for constructing line-items in FTW backend so that new pricing action can be used. This also adds some helper functions for handling lineItems and test setup for FTW backend.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,10 +94,11 @@
     "format-ci": "prettier --list-different '**/*.{js,css}'",
     "format-docs": "prettier --write '**/*.md'",
     "test": "NODE_ICU_DATA=node_modules/full-icu sharetribe-scripts test",
-    "test-ci": "sharetribe-scripts test --runInBand",
+    "test-ci": "yarn run test-server --runInBand && sharetribe-scripts test --runInBand",
     "eject": "sharetribe-scripts eject",
     "start": "node --icu-data-dir=node_modules/full-icu server/index.js",
     "dev-server": "export NODE_ENV=development PORT=4000 REACT_APP_CANONICAL_ROOT_URL=http://localhost:4000&&yarn run build&&nodemon --watch server server/index.js",
+    "test-server": "jest ./server/**/*.test.js",
     "heroku-postbuild": "yarn run build",
     "translate": "node scripts/translations.js"
   },
--- a/server/api-util/currency.js
+++ b/server/api-util/currency.js
@@ -49,6 +49,48 @@ exports.unitDivisor = currency => {
   return subUnitDivisors[currency];
 };
 
+////////// Currency manipulation in string format //////////
+
+/**
+ * Ensures that the given string uses only dots or commas
+ * e.g. ensureSeparator('9999999,99', false) // => '9999999.99'
+ *
+ * @param {String} str - string to be formatted
+ *
+ * @return {String} converted string
+ */
+const ensureSeparator = (str, useComma = false) => {
+  if (typeof str !== 'string') {
+    throw new TypeError('Parameter must be a string');
+  }
+  return useComma ? str.replace(/\./g, ',') : str.replace(/,/g, '.');
+};
+
+/**
+ * Ensures that the given string uses only dots
+ * (e.g. JavaScript floats use dots)
+ *
+ * @param {String} str - string to be formatted
+ *
+ * @return {String} converted string
+ */
+const ensureDotSeparator = str => {
+  return ensureSeparator(str, false);
+};
+
+/**
+ * Convert string to Decimal object (from Decimal.js math library)
+ * Handles both dots and commas as decimal separators
+ *
+ * @param {String} str - string to be converted
+ *
+ * @return {Decimal} numeral value
+ */
+const convertToDecimal = str => {
+  const dotFormattedStr = ensureDotSeparator(str);
+  return new Decimal(dotFormattedStr);
+};
+
 // Divisor can be positive value given as Decimal, Number, or String
 const convertDivisorToDecimal = divisor => {
   try {
--- a/server/api-util/currency.js
+++ b/server/api-util/currency.js
@@ -1,0 +1,139 @@
+const Decimal = require('decimal.js');
+const has = require('lodash/has');
+const { types } = require('sharetribe-flex-sdk');
+const { Money } = types;
+
+/** Helper functions for handling currency */
+
+// https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+// https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER
+// https://stackoverflow.com/questions/26380364/why-is-number-max-safe-integer-9-007-199-254-740-991-and-not-9-007-199-254-740-9
+const MIN_SAFE_INTEGER = Number.MIN_SAFE_INTEGER || -1 * (2 ** 53 - 1);
+const MAX_SAFE_INTEGER = Number.MAX_SAFE_INTEGER || 2 ** 53 - 1;
+
+const isSafeNumber = decimalValue => {
+  if (!(decimalValue instanceof Decimal)) {
+    throw new Error('Value must be a Decimal');
+  }
+  return decimalValue.gte(MIN_SAFE_INTEGER) && decimalValue.lte(MAX_SAFE_INTEGER);
+};
+
+// See: https://en.wikipedia.org/wiki/ISO_4217
+// See: https://stripe.com/docs/currencies
+const subUnitDivisors = {
+  AUD: 100,
+  CAD: 100,
+  CHF: 100,
+  CNY: 100,
+  DKK: 100,
+  EUR: 100,
+  GBP: 100,
+  HKD: 100,
+  INR: 100,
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -94,10 +94,11 @@
     "format-ci": "prettier --list-different '**/*.{js,css}'",
     "format-docs": "prettier --write '**/*.md'",
     "test": "NODE_ICU_DATA=node_modules/full-icu sharetribe-scripts test",
-    "test-ci": "sharetribe-scripts test --runInBand",
+    "test-ci": "yarn run test-server && sharetribe-scripts test --runInBand",
     "eject": "sharetribe-scripts eject",
     "start": "node --icu-data-dir=node_modules/full-icu server/index.js",
     "dev-server": "export NODE_ENV=development PORT=4000 REACT_APP_CANONICAL_ROOT_URL=http://localhost:4000&&yarn run build&&nodemon --watch server server/index.js",
+    "test-server": "jest ./server/**/*.test.js",
```
OtterleyW, line 101: CRA only supports automatically running tests inside `/src` so we need to add another script for server side tests. See e.g. https://spectrum.chat/create-react-app/general/how-can-tests-outside-of-src-be-included-in-the-test-run~3610ecf1-a396-466b-8f62-a62627062151

server/api-util/currency.js
```diff
@@ -0,0 +1,179 @@
...
+  NZD: 100,
+  SEK: 100,
+  SGD: 100,
+  USD: 100,
+};
+
+// Get the minor unit divisor for the given currency
+exports.unitDivisor = currency => {
+  if (!has(subUnitDivisors, currency)) {
+    throw new Error(`No minor unit divisor defined for currency: ${currency}`);
```
Gnito, line 45: Since this is about server-side tests and we have this subunitDivisor list duplicated, it might be good to mention the test environment in the error. "Check the allowed currencies for this test from..."

server/api-util/currency.js
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Ensures that the given string uses only dots or commas
+ * e.g. ensureSeparator('9999999,99', false) // => '9999999.99'
+ *
+ * @param {String} str - string to be formatted
+ *
+ * @return {String} converted string
+ */
+const ensureSeparator = (str, useComma = false) => {
```
Gnito, line 60: These "ensureSeparator" stuff were added due to users writing listing price in different localized number rules.
Is it likely that these server-side tests would get random input too? 
(These probably don't hurt the test, but they just obfuscate the code a bit.)

server/api-util/currency.js
```diff
@@ -0,0 +1,179 @@
...
+ */
+exports.convertMoneyToNumber = value => {
+  if (!(value instanceof Money)) {
+    throw new Error('Value must be a Money type');
+  }
+  const subUnitDivisorAsDecimal = convertDivisorToDecimal(this.unitDivisor(value.currency));
+  let amount;
+
+  if (isGoogleMathLong(value.amount)) {
+    // TODO: temporarily also handle goog.math.Long values created by
```
Gnito, line 160: I'm not sure if this is still valid. It could be that this is just a forgotten piece of code.

server/api-util/currency.test.js
```diff
@@ -0,0 +1,57 @@
...
+      expect(() => convertUnitToSubUnit(1, 'asdf')).toThrowError();
+    });
+  });
+
+  describe('convertMoneyToNumber(value)', () => {
+    it('Money as value', () => {
+      expect(convertMoneyToNumber(new Money(10, 'USD'))).toBeCloseTo(0.1);
+      expect(convertMoneyToNumber(new Money(1000, 'USD'))).toBeCloseTo(10);
+      expect(convertMoneyToNumber(new Money(9900, 'USD'))).toBeCloseTo(99);
+      expect(convertMoneyToNumber(new Money(10099, 'USD'))).toBeCloseTo(100.99);
```
Gnito, line 44: Uh, usage of `toBeCloseTo` sounds strange. This might be a buggy code originally. 

In plain JS (and underlying C), `0.1 + 0.2 != 0.3` due to rounding of floats, but it sounds bad if `convertMoneyToNumber(new Money(10099, 'USD'))` is not equal to 100.99. (That's the reason why we use Decimal library and not plain JS, when calculating with money.)

server/api-util/lineItems.js
```diff
@@ -0,0 +1,53 @@
+const {
+  calculateQuantityFromDates,
+  calculateTotalFromLineItems,
+  consrtuctValidLineItems,
```
Gnito, line 4: Typo "construct"

Gnito: It seems that this function is not used in this file (Still it's best to fix the typo elsewhere)

server/api-util/lineItemHelpers.js
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Calculates lineTotal for lineItem based on percentage.
+ * The total will be `unitPrice * (percentage / 100)`.
+ *
+ * @param {Money} unitPrice
+ * @param {int} percentage
+ *
+ * @returns {Money} lineTotal
+ */
+exports.calculateTotalPriceFromPercentage = (unitPrice, percentage) => {
```
Gnito, line 40: Can the percentage be anything? e.g. -200, 345?

server/api-util/lineItemHelpers.js
```diff
@@ -0,0 +1,192 @@
...
+ * Calculates lineTotal for lineItem based on seats and units.
+ * The total will be `unitPrice * units * seats`.
+ *
+ * @param {Money} unitPrice
+ * @param {int} unitCount
+ * @param {int} seats
+ *
+ * @returns {Money} lineTotal
+ */
+exports.calculateTotalPriceFromSeats = (unitPrice, unitCount, seats) => {
```
Gnito, line 62: Can any of the params (unitPrice, unitCount or seats) be negative?

server/api-util/lineItemHelpers.js
```diff
@@ -0,0 +1,192 @@
...
+ * @retuns {Money} total sum
+ */
+exports.calculateTotalFromLineItems = lineItems => {
+  const numericTotalPrice = lineItems.reduce((sum, lineItem) => {
+    const lineTotal = this.calculateLineTotal(lineItem);
+    const numericPrice = convertMoneyToNumber(lineTotal);
+    return new Decimal(numericPrice).add(sum);
+  }, 0);
+
+  const unitPrice = lineItems[0].unitPrice;
```
Gnito, line 132: I think there should be a check that we are checking correct line-item (I guess "line-item/night") and not trust that first line-item is always pointing to that instead of being a "line-item/provider-commission" or something.

Gnito: D'oh, never mind. this was just about checking currency!

server/api-util/currency.js
OtterleyW: We are not expecting random input because the value comes always form Money object so I guess there is no need to test this.

server/api-util/currency.js
OtterleyW: How can I check if this is still needed or not? I think I have never seen the warning and I tried to check the SDK but I'm not sure how to know if this can be removed or not...

server/api-util/currency.test.js
OtterleyW: Yeah, these tests have been copied from frontend so the issue is also there. I'm going to fix these in both places.

server/api-util/lineItemHelpers.js
OtterleyW: I think in theory it could be. On API side the limitations come from payin and payout sums which can't be negative and payin must be greater or equal to payout. If the percentage is calculated from just one lineItem (e.g. cleaning fee or something else small) even that big percentage might work inside those payin and payout limits.

server/api-util/lineItemHelpers.js
OtterleyW: As far as I understand from specs here (https://github.com/sharetribe/sharetribe-custom/blob/master/src/sharetribe/tegel/action/privileged_set_line_items.clj#L176) unitPrice and unitCount can be also negative but seats can't (maybe @kpuputti or @sktoiva can confirm). So maybe we should check that at least seats is not a negative value.

package.json
kpuputti: You probably should add the [--runInBand](https://jestjs.io/docs/en/cli.html#--runinband) option also for the server Jest command. Circle CI at least used to have lots of performance problems without that option.

server/api-util/lineItems.js
```diff
@@ -0,0 +1,47 @@
...
+ * Line item must have either `quantity` or `percentage` or both `seats` and `units`.
+ *
+ * `includeFor` defines commissions. Customer commission is added by defining `includeFor` array `["customer"]` and provider commission by `["provider"]`.
+ *
+ * @param {Object} listing
+ * @param {Object} bookingData
+ * @returns {Array} lineItems
+ */
+exports.transactionLineItems = (listing, bookingData) => {
+  const unitPrice = listing.attributes.unitPrice;
```
kpuputti, line 27: Now that I got to testing this, I notice that listings don't have a `unitPrice` attribute. It should be just `price`.

```suggestion
  const unitPrice = listing.attributes.price;
```

PR #1305, opened 2020-06-09
Use lineItems endpoint to in estimated transaction

OtterleyW:
Fetch lineItems form FTW backend to EstimatedTransactionMaybe component on ListingPage so that also unauthenticated users will see BookingBreakdown.

Kapture 2020-06-11 at 14 37 51

### DIFF
--- a/src/forms/BookingDatesForm/BookingDatesForm.js
+++ b/src/forms/BookingDatesForm/BookingDatesForm.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
-import { string, bool, arrayOf } from 'prop-types';
+import { string, bool, arrayOf, array, func } from 'prop-types';
 import { compose } from 'redux';
-import { Form as FinalForm } from 'react-final-form';
+import { Form as FinalForm, FormSpy } from 'react-final-form';
 import { FormattedMessage, intlShape, injectIntl } from '../../util/reactIntl';
 import classNames from 'classnames';
 import moment from 'moment';
@@ -9,7 +9,7 @@ import { required, bookingDatesRequired, composeValidators } from '../../util/va
 import { START_DATE, END_DATE } from '../../util/dates';
 import { propTypes } from '../../util/types';
 import config from '../../config';
-import { Form, PrimaryButton, FieldDateRangeInput } from '../../components';
+import { Form, IconSpinner, PrimaryButton, FieldDateRangeInput } from '../../components';
 import EstimatedBreakdownMaybe from './EstimatedBreakdownMaybe';
 
 import css from './BookingDatesForm.css';
@@ -22,6 +22,7 @@ export class BookingDatesFormComponent extends Component {
     this.state = { focusedInput: null };
     this.handleFormSubmit = this.handleFormSubmit.bind(this);
     this.onFocusedInputChange = this.onFocusedInputChange.bind(this);
+    this.handleOnChange = this.handleOnChange.bind(this);
   }
 
   // Function that can be passed to nested components
@@ -44,6 +45,25 @@ export class BookingDatesFormComponent extends Component {
       this.setState({ focusedInput: END_DATE });
     } else {
       this.props.onSubmit(e);
+    }
+  }
+
+  // When the values of the form are updated we need to fetch
+  // lineItems from FTW backend for the EstimatedTransactionMaybe
+  // In case you add more fields to the form, make sure you add
+  // the values here to the bookingData object.
+  handleOnChange(formValues) {
+    const { startDate, endDate } =
+      formValues.values && formValues.values.bookingDates ? formValues.values.bookingDates : {};
+    const listingId = this.props.listingId;
+    const isOwnListing = this.props.isOwnListing;
+
+    if (startDate && endDate && !this.props.fetchLineItemsInProgress) {
+      this.props.onFetchTransactionLineItems({
+        bookingData: { startDate, endDate },
+        listingId,
+        isOwnListing,
+      });
     }
   }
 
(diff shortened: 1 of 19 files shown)


server/apiServer.js
```diff
@@ -8,6 +8,7 @@ const express = require('express');
 const cookieParser = require('cookie-parser');
 const bodyParser = require('body-parser');
 const cors = require('cors');
+const sharetribeSdk = require('sharetribe-flex-sdk');
```
kpuputti, line 11: Was this import a leftover from fixing conflicts from rebase? No need to import the SDK here anymore.

src/forms/BookingDatesForm/BookingDatesForm.js
```diff
@@ -47,6 +48,25 @@ export class BookingDatesFormComponent extends Component {
...
+  // lineItems from FTW backend for the EstimatedTransactionMaybe
+  // In case you add more fields to the form, make sure you add
+  // the values here to the bookingData object.
+  handleOnChange(formValues) {
+    const { startDate, endDate } =
+      formValues.values && formValues.values.bookingDates ? formValues.values.bookingDates : {};
+    const listingId = this.props.listingId;
+    const isOwnListing = this.props.isOwnListing;
+
+    if (startDate && endDate && !this.props.fetchLineItemsInProgress) {
```
kpuputti, line 61: Is it possible to change the values in the form while the line items are being fetched?

If a request takes a long time, and the user can change the dates, the line items won't be fetched for the new dates. In that case the fetch for the latter dates is ignored, and the UI can be in a weird state.

If there isn't a simple solution to cancel/ignore the previous request when a new is being made, then the UI can always prevent selecting new values when the line items are being fetched. Or is it already doing just that?

src/forms/BookingDatesForm/BookingDatesForm.js
```diff
@@ -84,54 +104,67 @@ export class BookingDatesFormComponent extends Component {
...
+          const bookingInfo =
+            bookingData && lineItems && !fetchLineItemsError ? (
+              <div className={css.priceBreakdownContainer}>
+                <h3 className={css.priceBreakdownTitle}>
+                  <FormattedMessage id="BookingDatesForm.priceBreakdownTitle" />
+                </h3>
+                <EstimatedBreakdownMaybe bookingData={bookingData} lineItems={lineItems} />
+              </div>
+            ) : fetchLineItemsInProgress ? (
+              <IconSpinner className={css.spinner} />
```
kpuputti, line 161: Hmm, this nested ternary is a bit hard to understand. I know it's already quite common in the FTW codebase, but I'm not sure if `bookingData && lineItems && !fetchLineItemsError` and `fetchLineItemsInProgress` can both be true. Then the estimation would be rendered since it's checked first, even if the fetch is in progress.

This is related to the in-progress fetch order mentioned in the previous comment.

src/forms/BookingDatesForm/BookingDatesForm.test.js
```diff
@@ -13,6 +13,16 @@ import EstimatedBreakdownMaybe from './EstimatedBreakdownMaybe';
 const { Money } = sdkTypes;
 
 const noop = () => null;
+const lineItems = [
+  {
+    code: 'line-item/nights',
```
kpuputti, line 18: Should this be changed? At least the other code uses singular form.

```suggestion
    code: 'line-item/night',
```

src/forms/BookingDatesForm/EstimatedBreakdownMaybe.js
```diff
@@ -29,40 +29,44 @@ import React from 'react';
...
-const estimatedTotalPrice = (unitPrice, unitCount) => {
-  const numericPrice = convertMoneyToNumber(unitPrice);
-  const numericTotalPrice = new Decimal(numericPrice).times(unitCount).toNumber();
+const estimatedTotalPrice = lineItems => {
+  const numericTotalPrice = lineItems.reduce((sum, lineItem) => {
+    const numericPrice = convertMoneyToNumber(lineItem.lineTotal);
+    return new Decimal(numericPrice).add(sum);
+  }, 0);
+
+  const unitPrice = lineItems[0].unitPrice;
```
kpuputti, line 48: Is it guaranteed that there always is at least a single line item? I guess a customization could break this.

src/forms/BookingDatesForm/EstimatedBreakdownMaybe.js
```diff
@@ -87,18 +91,9 @@ const estimatedTransaction = (unitType, bookingStart, bookingEnd, unitPrice, qua
...
-          includeFor: ['customer', 'provider'],
-          unitPrice: unitPrice,
-          quantity: new Decimal(unitCount),
-          lineTotal: totalPrice,
-          reversal: false,
-        },
-      ],
+      payinTotal,
+      payoutTotal,
+      lineItems: customerLineItems,
```
kpuputti, line 96: Before this component was for both customer and provide, although it's only ever shown for customers in the listing page. However, now the customer line items are only used, so maybe at least add a clear comment that the estimation breakdown is for the customer. Possibly even rename the component?

server/apiServer.js
OtterleyW: Yeah there was quite a lot of merge conflicts so I think this is just leftover, I will remove it

src/forms/BookingDatesForm/EstimatedBreakdownMaybe.js
OtterleyW: I don't know if it's possible that there is e.g. no provider lineItems at all but if that's the case then this would be a problem. I will change this so that we use marketplace currency from config here as a default if there are no lineItems (so that we can return 0 to e.g. payout sum).

src/forms/BookingDatesForm/EstimatedBreakdownMaybe.js
OtterleyW: I decided to add customerRole variable there so that it would be easier to use the same component also for the provider if needed in the future or in some customization.

src/forms/BookingDatesForm/BookingDatesForm.js
OtterleyW: Good catch! I wasn't sure about an easy way to cancel/ignore the previous request so I decided to go with disabling the form while fetch is in progress.

PR #1310, opened 2020-06-15
Flexible pricing UI

kpuputti:
(This PR combines #1305 and #1309 as well as new changes.)

This PR adds opt-in support for flexible pricing and privileged transitions. Privileged transitions are transaction process transitions that need to be run from a secure context, i.e. the backend.

The opt-in UI support for privileged transitions adds integration to the FTW API endpoints added in #1301 .

## Transaction line items

The line items that define the pricing of a transaction are now defined securely in the backend. This means that privileged transitions can safely define the line items in the backend based on the booking data, and the pricing can be fully customized in FTW. This means changing commissions, adding extra line items for discounts, etc.

### Important note

The `EstimatedBreakdownMaybe` component that is used in the `ListingPage` will no longer internally define the pricing line items. Instead, the `/api/transaction-line-items` endpoint will be used to fetch the line items.

This means that the same place that defines the line items for privileged transitions is also used for the price estimation. This also means that any pricing customizations will need to be migrated to the `transactionLineItems` function in `server/api-util/lineItems.js` that will become the main customization point for describing pricing when privileged transitions are the default.

## Using privileged transitions

If you want to opt-in to marking some of the transitions as privileged, you can edit the `isPrivileged` function in `src/util/transaction.js` and add the transitions there. After that, those transitions will automatically go through the backend.

Note that we will provide more specific documentation for this when the privileged transitions become the default. Also note that using privileged transitions requires a process change with Flex CLI.

## Customizing pricing

As mentioned earlier, the `transactionLineItems` function in `server/api-util/lineItems.js` will become the main customization point to describe the pricing in a transaction.

## TODO

- [x] Call the local privileged transition endpoint
- [x] Make sure all booking breakdowns use the new line items
- [x] Call speculate transition (not initiate) when moving from enquiry
- [x] Make privileged endpoints optional based on tx config
- [x] Ensure correct tx process name in config
- [x] Make sure server side works
- [x] Update changelog

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -32,7 +32,7 @@ const sortSearchByDistance = false;
 //
 // In a way, 'processAlias' defines which transaction process (or processes)
 // this particular web application is able to handle.
-const bookingProcessAlias = 'preauth-nightly-booking/release-1';
+const bookingProcessAlias = 'privileged-custom-pricing/release-1';
 
 // The transaction line item code for the main unit type in bookings.
 //
--- a/src/config.js
+++ b/src/config.js
@@ -32,7 +32,7 @@ const sortSearchByDistance = false;
 //
 // In a way, 'processAlias' defines which transaction process (or processes)
 // this particular web application is able to handle.
-const bookingProcessAlias = 'privileged-custom-pricing/release-1';
+const bookingProcessAlias = 'preauth-nightly-booking/release-1';
 
 // The transaction line item code for the main unit type in bookings.
 //
--- a/src/containers/CheckoutPage/CheckoutPage.duck.js
+++ b/src/containers/CheckoutPage/CheckoutPage.duck.js
@@ -1,5 +1,6 @@
 import pick from 'lodash/pick';
 import config from '../../config';
+import { initiatePrivileged, transitionPrivileged } from '../../util/api';
 import { denormalisedResponseEntities } from '../../util/data';
 import { storableError } from '../../util/errors';
 import {
@@ -161,6 +162,12 @@ export const stripeCustomerError = e => ({
 
 export const initiateOrder = (orderParams, transactionId) => (dispatch, getState, sdk) => {
   dispatch(initiateOrderRequest());
+
+  const bookingData = {
+    startDate: orderParams.bookingStart,
+    endDate: orderParams.bookingEnd,
+  };
+
   const bodyParams = transactionId
     ? {
         id: transactionId,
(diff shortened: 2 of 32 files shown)


src/config.js
```diff
@@ -32,7 +32,7 @@ const sortSearchByDistance = false;
 //
 // In a way, 'processAlias' defines which transaction process (or processes)
 // this particular web application is able to handle.
-const bookingProcessAlias = 'preauth-nightly-booking/release-1';
+const bookingProcessAlias = 'privileged-custom-pricing/release-1';
```
ovan, line 35: Is this the new process name we've chosen @sktoiva? It puts a lot of emphasis on the privileged + custom pricing parts. Those are the new thing so it makes sense in this moment of time but if we'd come out the gate with this process then would we call it that or instead just something like flex-default-process/release-1? There's no more one defining characteristic for this process like before when it was the pricing. Now it's just a default starting point that supports many kind of pricing structures but also encodes things like double blind reviews and some choices about possibility to cancel and notification emails. We won't get too many chances to rename the process so worth thinking this through. I'd like to hear your thoughts before we commit @sktoiva @kpuputti @lyyder.

src/containers/CheckoutPage/CheckoutPage.duck.js
```diff
@@ -177,9 +184,34 @@ export const initiateOrder = (orderParams, transactionId) => (dispatch, getState
...
+        dispatch(initiateOrderError(storableError(e)));
+        const transactionIdMaybe = transactionId ? { transactionId: transactionId.uuid } : {};
+        log.error(e, 'initiate-order-failed', {
+          ...transactionIdMaybe,
+          listingId: orderParams.listingId.uuid,
+          bookingStart: orderParams.bookingStart,
+          bookingEnd: orderParams.bookingEnd,
+        });
+        throw e;
+      });
```
ovan, line 206: This looks like the same code as repeated below. Extract to error handler function?

src/containers/CheckoutPage/CheckoutPage.duck.js
```diff
@@ -177,9 +184,34 @@ export const initiateOrder = (orderParams, transactionId) => (dispatch, getState
...
+  }
 
-  return createOrder(bodyParams, queryParams)
+  return transitionPrivileged({
+    isSpeculative: false,
+    bookingData,
+    bodyParams,
+    queryParams,
+  })
```
ovan, line 214: This might the style that is used in FTW widely, in which case you can just ignore my comment, but I find the early return in if, followed by implicit else that also returns, very confusing to read. So technically else is not needed here because if returns but for clarity I'd put this in an else block.

src/config.js
kpuputti: Oh, I should've added a comment here. This is just the name I put to the new process in our staging Saunatime marketplace. We will need to decide the new process name here or later, but This PR will most likely undo this change and release a version of FTW that uses the new line items endpoint for the estimation, but otherwise keep the current default process with its non-privileged transitions.

Then we can easily enable privileged transitions by marking some of the transitions in `util/transactions.js` as privileged, and at some point make that the default.

Does this make sense?

src/containers/CheckoutPage/CheckoutPage.duck.js
kpuputti: Yeah, this is one of those things that yield debate. We used to have a lint rule that enforced one over the other (don't remember which), but at least at some point I think Prettier rewrote some of those else statements away.

src/config.js
kpuputti: Summary: the new process will be `flex-default-process`, but that will be taken into use in another PR. Core is already using that for recent marketplaces that have the process with privileged transitions added.

PR #1312, opened 2020-06-23
Add rounding to privilate-line-item price estimations

Gnito:
and remove unnecessary copy-pasted code. 
(Web app has functions that deal with user inputs - privileged transitions on server-side don't need to deal with those.)

### DIFF
--- a/server/api-util/lineItemHelpers.js
+++ b/server/api-util/lineItemHelpers.js
@@ -3,7 +3,7 @@ const has = require('lodash/has');
 const { types } = require('sharetribe-flex-sdk');
 const { Money } = types;
 
-const { convertMoneyToNumber, unitDivisor, convertUnitToSubUnit } = require('./currency');
+const { getAmountAsDecimalJS, convertDecimalJSToNumber } = require('./currency');
 const { nightsBetween, daysBetween } = require('./dates');
 const LINE_ITEM_NIGHT = 'line-item/night';
 const LINE_ITEM_DAY = 'line-item/day';
@@ -20,12 +20,15 @@ const LINE_ITEM_DAY = 'line-item/day';
  * @returns {Money} lineTotal
  */
 exports.calculateTotalPriceFromQuantity = (unitPrice, unitCount) => {
-  const numericPrice = convertMoneyToNumber(unitPrice);
-  const numericTotalPrice = new Decimal(numericPrice).times(unitCount).toNumber();
-  return new Money(
-    convertUnitToSubUnit(numericTotalPrice, unitDivisor(unitPrice.currency)),
-    unitPrice.currency
-  );
+  const amountFromUnitPrice = getAmountAsDecimalJS(unitPrice);
+
+  // NOTE: We round the total price to the nearest integer.
+  //       Payment processors don't support fractional subunits.
+  const totalPrice = amountFromUnitPrice.times(unitCount).toNearest(1, Decimal.ROUND_HALF_UP);
+  // Get total price as Number (and validate that the conversion is safe)
+  const numericTotalPrice = convertDecimalJSToNumber(totalPrice);
+
+  return new Money(numericTotalPrice, unitPrice.currency);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


server/api-util/lineItemHelpers.js
```diff
@@ -38,15 +38,15 @@ exports.calculateTotalPriceFromQuantity = (unitPrice, unitCount) => {
...
+  const amountFromUnitPrice = getAmountAsDecimalJS(unitPrice);
+  const totalPrice = amountFromUnitPrice
     .times(percentage)
     .dividedBy(100)
-    .toNumber();
-  return new Money(
-    convertUnitToSubUnit(numericTotalPrice, unitDivisor(unitPrice.currency)),
-    unitPrice.currency
-  );
+    .round();
```
ovan, line 45: I can see that the default rounding mode in decimal.js is half up (https://mikemcl.github.io/decimal.js/#rounding) but I'm wondering if that's a stable default we should just trust to be there. It would introduce a hard to spot bug if some future version would change the default mode. That might be unlikely but explicitly setting the rounding mode would eliminate that possibility. Another reason to be explicit is to document it clearly that that's what we're doing. Maybe that could be set in `getAmountAsDecimalJS`?

server/api-util/lineItemHelpers.js
```diff
@@ -20,12 +20,12 @@ const LINE_ITEM_DAY = 'line-item/day';
...
  */
 exports.calculateTotalPriceFromQuantity = (unitPrice, unitCount) => {
-  const numericPrice = convertMoneyToNumber(unitPrice);
-  const numericTotalPrice = new Decimal(numericPrice).times(unitCount).toNumber();
-  return new Money(
-    convertUnitToSubUnit(numericTotalPrice, unitDivisor(unitPrice.currency)),
-    unitPrice.currency
-  );
+  const amountFromUnitPrice = getAmountAsDecimalJS(unitPrice);
+  const totalPrice = amountFromUnitPrice.times(unitCount);
```
ovan, line 24: I think we need rounding here too as unitCount (which I assume is the same as quantity) can be a non-integer

server/api-util/lineItemHelpers.js
```diff
@@ -63,15 +63,13 @@ exports.calculateTotalPriceFromSeats = (unitPrice, unitCount, seats) => {
...
-    .times(unitCount)
-    .times(seats)
-    .toNumber();
-  return new Money(
-    convertUnitToSubUnit(numericTotalPrice, unitDivisor(unitPrice.currency)),
-    unitPrice.currency
-  );
+
+  const amountFromUnitPrice = getAmountAsDecimalJS(unitPrice);
+  const totalPrice = amountFromUnitPrice.times(unitCount).times(seats);
```
ovan, line 68: I think we need rounding here too as both unitCount (which I assume is the same as units) and seats can be non-integers.

server/api-util/lineItemHelpers.js
Gnito: @ovan I'm not sure if that is something we should be afraid of - the whole math library commits suicide if they change their default rounding mode. 

Anyway, `round()` doesn't take rounding mode as an optional parameter. It uses a library-wide variable for that - and I don't want to hard-code that variable in a random file. 
So, I'll change this to `.toNearest(1, Decimal.ROUND_HALF_UP)`.

Gnito: I added these now here. `getAmountAsDecimalJS` is called before the calculations, so it's not the right place. And to be frank, I like to remind people that there are constraints that they need to be aware of.

So, the code now looks like this:
```js
  // NOTE: We round the amount to the nearest integer.
  //       Payment processors don't support fractional subunits.
  const totalPrice = amountFromUnitPrice
    .times(percentage)
    .dividedBy(100)
    .toNearest(1, Decimal.ROUND_HALF_UP);
```

server/api-util/lineItemHelpers.js
```diff
@@ -20,12 +20,15 @@ const LINE_ITEM_DAY = 'line-item/day';
...
-  const numericTotalPrice = new Decimal(numericPrice).times(unitCount).toNumber();
-  return new Money(
-    convertUnitToSubUnit(numericTotalPrice, unitDivisor(unitPrice.currency)),
-    unitPrice.currency
-  );
+  const amountFromUnitPrice = getAmountAsDecimalJS(unitPrice);
+
+  // NOTE: We round the total price to the nearest integer.
+  //       Payment processors don't support fractional subunits.
+  const totalPrice = amountFromUnitPrice.times(unitCount).toNearest(1, Decimal.ROUND_HALF_UP);
```
sktoiva, line 27: You're the best to comment on style issues, but should these be on newlines as below?

server/api-util/lineItemHelpers.js
ovan: @Gnito I think that's perfect. I didn't realize `toNeareast` exists and was only looking at `round` which didn't take rounding mode as a parameter. Great that this one does and the rounding mode is now clearly visible here. 👍

server/api-util/lineItemHelpers.js
Gnito: @sktoiva Prettier...

PR #1314, opened 2020-06-25
Opt-out flexible pricing

lyyder:
Based on #1310.

Make new Flex pricing the default pricing method.

- Add price calculation transitions to the list of privileged transitions
- Update transaction process definition
- Update default process alias


lyyder:
@kpuputti indeed

PR #1317, opened 2020-06-29
BookingUnitType config.js variable needed more comments

Gnito:
This adds more comments about how to change booking unit type. 
I.e. the new API server/proxy creates it a bit more work for people who want to change the unit type from ´line-item/night´ to ´line-item/day´. 

We could also pass the bookingUnitType variable among bookingData to client app's server for flexible pricing calculations, but 
1) somehow adding more parameters that affect price calculations feels a bit wrong to me (customizers should think security concerns twice if they add more parameters)
2) "bookingData" has a bit different meaning in the web app's context (it contains any extra info from BookingDatesForm).

### DIFF
--- a/server/api-util/lineItems.js
+++ b/server/api-util/lineItems.js
@@ -2,7 +2,9 @@ const { calculateQuantityFromDates, calculateTotalFromLineItems } = require('./l
 const { types } = require('sharetribe-flex-sdk');
 const { Money } = types;
 
-const unitType = 'line-item/night';
+// This bookingUnitType needs to be one of the following:
+// line-item/night, line-item/day or line-item/units
+const bookingUnitType = 'line-item/night';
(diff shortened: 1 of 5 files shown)


server/api-util/lineItems.js
```diff
@@ -2,7 +2,9 @@ const { calculateQuantityFromDates, calculateTotalFromLineItems } = require('./l
 const { types } = require('sharetribe-flex-sdk');
 const { Money } = types;
 
-const unitType = 'line-item/night';
+// This bookingUnitType needs to be one of the following:
+// line-item/night, line-item/day or line-item/units
+const bookingUnitType = 'line-item/night';
```
Gnito, line 7: I changed the name so that it's in sync with the config.js variable (in web app).

PR #1318, opened 2020-06-29
Fix typo

Gnito:
SET_INITAL_VALUES => SET_INITIAL_VALUES


Gnito:
This fix originally comes from FTW-hourly: https://github.com/sharetribe/ftw-hourly/pull/100
But let's fix it first here and then push it to FTW-hourly in the next release.

PR #1319, opened 2020-06-29
Fix typo

Gnito:


Gnito:
This fix originally comes from FTW-hourly: sharetribe/ftw-hourly#102
But let's fix it first here and then push it to FTW-hourly in the next release.

PR #1334, opened 2020-07-31
Update en.json

BookOBO:


OtterleyW:
This seems to be accidental PR from customization project so I'm closing it.

PR #1342, opened 2020-08-12
Custom avatar image

lyyder:
Takes Flex's custom image variants functionality into use by requesting two custom variants for the avatar. The variant sizes are 40x40 and 80x80.

### DIFF
--- a/src/components/Avatar/Avatar.js
+++ b/src/components/Avatar/Avatar.js
@@ -19,6 +19,12 @@ const AVATAR_SIZES_MEDIUM = '60px';
 const AVATAR_SIZES_LARGE = '96px';
 
 const AVATAR_IMAGE_VARIANTS = [
+  // 40x40
+  'square-xsmall',
+
+  // 80x80
+  'square-xsmall2x',
+
   // 240x240
   'square-small',
 
--- a/src/ducks/user.duck.js
+++ b/src/ducks/user.duck.js
@@ -5,6 +5,7 @@ import { LISTING_STATE_DRAFT } from '../util/types';
 import * as log from '../util/log';
 import { authInfo } from './Auth.duck';
 import { stripeAccountCreateSuccess } from './stripeConnectAccount.duck';
+import { util as sdkUtil } from '../util/sdkLoader';
 
 // ================ Action types ================ //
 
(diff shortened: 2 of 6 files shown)


src/ducks/user.duck.js
```diff
@@ -317,7 +318,17 @@ export const fetchCurrentUser = (params = null) => (dispatch, getState, sdk) =>
 
   const parameters = params || {
     include: ['profileImage', 'stripeAccount'],
-    'fields.image': ['variants.square-small', 'variants.square-small2x'],
+    'fields.image': ['variants.avatar', 'variants.avatar2x'],
```
Gnito, line 321: Since this is in user.duck.js, does this affect the ProfileSettingsPage and ProfilePage too? I thought that you would only add new variants.

src/components/Avatar/Avatar.js
```diff
@@ -19,11 +19,11 @@ const AVATAR_SIZES_MEDIUM = '60px';
 const AVATAR_SIZES_LARGE = '96px';
```
OtterleyW, line 19: Is the 80x80 avatar image variant large enough for large avatar component which seems to be 96x96? Earlier it has used the 240x240 variant

Gnito: @OtterleyW good question. 96px avatar is used at least on listing page, where it is actually author's avatar and therefore fetched separately. If that's the only place, then it doesn't affect currentUser's avatar sizes.

Gnito: Hmm. TopbarMobileMenu uses it too -> big retina version should be left there since a high pixel density is what you expect in mobile phones. (240px aka square-small should be fine for AvatarLarge component)

src/ducks/user.duck.js
lyyder: Indeed, is the square-small2x needed? Would just adding square-small to that list do?

src/components/Avatar/Avatar.js
lyyder: True, would it make sense to have the variants 40x40, 100x100, and 240x240 (square-small)?

src/ducks/user.duck.js
Gnito: @lyyder if ProfileSettingPage gets the image from user.duck.js, then this needs also retina version (i.e. both 'variants.square-small', 'variants.square-small2x'). 

So, if that's the case, then I'd suggest that you just add those new avatar variants (for Topbar) as extra. The components & browser then decide which of the image they are going to use.

src/components/Avatar/Avatar.js
Gnito: There are two parallel threads that discuss the same issue. My suggestion is that you just add 40x40 and 80x80 and keep the current ones there too ('variants.square-small', 'variants.square-small2x') because ProfileSettingsPage needs them.

PR #1343, opened 2020-08-12
Fix warning

Gnito:


Gnito:
Added Changelog too to
https://github.com/sharetribe/ftw-daily/pull/1340

PR #1347, opened 2020-08-13
Update helmet to v4

OtterleyW:
In Helmet v4 there were some changes to content security policy, see: https://github.com/helmetjs/helmet/wiki/Helmet-4-upgrade-guide#what-changed-in-the-content-security-policy-middleware

In addition to those changes mentioned in the documentation it seems that Helmet is now expecting every directive to be iterable (https://github.com/helmetjs/helmet/blob/v4.0.0/middlewares/content-security-policy/index.ts) We have used also string and boolean values in directives which were causing error after the update and needed to be changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,16 +14,20 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2020-XX-XX
 
+- [change] Update Helmet to v4.0.0. Show warning if environment variable REACT_APP_CSP is not set or
+  if it's set to 'report' mode in production environmet. Set REACT_APP_CSP to 'report' mode by
+  default in `.env-template` file. [#1347](https://github.com/sharetribe/ftw-daily/pull/1347)
 - [change] In `StripeConnectAccountForm` show error message from Stripe if there is one when
   fetching account link. [#1346](https://github.com/sharetribe/ftw-daily/pull/1346)
 
 ## [v6.2.0] 2020-08-12
 
-This change set was originally released as a patch update 6.1.2 but after
-reconsideration it's released as a minor update 6.2.0.
+This change set was originally released as a patch update 6.1.2 but after reconsideration it's
+released as a minor update 6.2.0.
 
 - [fix] remove typo [#1343](https://github.com/sharetribe/ftw-daily/pull/1343)
-- [change] Request custom image variants for avatar [#1342](https://github.com/sharetribe/ftw-daily/pull/1342)
+- [change] Request custom image variants for avatar
+  [#1342](https://github.com/sharetribe/ftw-daily/pull/1342)
 - [change] Some dependency updates [#1337](https://github.com/sharetribe/ftw-daily/pull/1337)
 - [fix] Use Stripe's `confirmCardPayment` function instead of deprecated `handleCardPayment` to
   confirm PaymentIntent. In addition to the rename, the arguments passed to `handleCardPayment` are
--- a/server/index.js
+++ b/server/index.js
@@ -69,7 +69,14 @@ app.use(log.requestHandler());
 
 // The helmet middleware sets various HTTP headers to improve security.
 // See: https://www.npmjs.com/package/helmet
-app.use(helmet());
+// Helmet 4 doesn't disable CSP by default so we need to do that explicitly.
+// If csp is enabled we will add that separately.
+
+app.use(
+  helmet({
+    contentSecurityPolicy: false,
+  })
+);
 
 if (cspEnabled) {
   // When a CSP directive is violated, the browser posts a JSON body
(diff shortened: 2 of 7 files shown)


server/index.js
```diff
@@ -69,7 +69,14 @@ app.use(log.requestHandler());
 
 // The helmet middleware sets various HTTP headers to improve security.
 // See: https://www.npmjs.com/package/helmet
-app.use(helmet());
+// Helmet 4 doesn't disable CSP by default so we need to do that explicitly.
+// If csp is enabled we will add that separately.
+
+app.use(
+  helmet({
+    contentSecurityPolicy: false,
```
Gnito, line 77: @OtterleyW This made me wonder if we should change our default too? I'm afraid that people don't really turn CSP on and that's pretty insecure setup. How understandable error there is if 'block' is on by default? 

I guess the minimum change would be to just set REACT_APP_CSP to 'block' in .env-template and see that it goes through in config script.

@kpuputti you made this setup originally, do you have opinion about this?

Gnito: Or if that's too much, we could add console warning that "You don't have CSP turned on. You should turn it on before going to production"

kpuputti: I don't have strong opinions here as I don't really know how much people use CSP and how important they think it is.

One thing to not is that CSP only works when you run the app with a server, the usual `yarn run dev` doesn't have any CSP features on, so locally you won't see any reports or blocked requests when developing. This makes the `.env-template` default value quite useless IMO as that file is not used in deployments.

So I'm not sure how much we can help by default. I think report mode by default in production should be fine, as the issues might be difficult to debug and only happen with actual users. A generic CSP warning is also ok, but the reports already show up in the dev tools console.

OtterleyW: @Gnito Do you mean that we should add our own csp as default (meaning that the mode would be report or block) or that we should allow Helmet to add the default csp mode if nothing else is specified (you can see the content here https://github.com/helmetjs/helmet/blob/v4.0.0/middlewares/content-security-policy/index.ts#L48)? 

We have default value for csp in app.json if the deploy button is used for deploying to Heroku https://github.com/sharetribe/ftw-daily/blob/master/app.json#L42 
Otherwise, we can't affect much on what env variables people use in their prod environments.

Gnito: @OtterleyW own default or warning. e.g. stripe.js has some warnings in console log. Not directly related to this piece of code (unless we decide to add console.warn here.

@kpuputti yes, but it would show when ppl use config script or test SSR with ”yarn run dev-server”

Gnito: @kpuputti reports don’t show if ppl don’t use CSP - or am I wrong here? 

I’m talking about adding warning if REACT_APP_CSP is not set at all. Or if it is in ”report” on production environment. There’s two possibility to set that warning: node.js console.warn or Web Inspector’s  console.

kpuputti: Ah, I see. Indeed by default the server has the CSP disable if the env var is missing. So no reports will be seen in a deployment without the env var.

It sounds like defaulting to "report" or adding a warning makes sense. Warning might be nicer as the reports can fill up the console output.

kpuputti: And by warning I mean a warning in the frontend console. Having it in the server logs is easily ignored.

Gnito: I'm not sure if "report" would generate that much log. At least, we don't get many reports to Sentry. So, I assume that the issue would only be manifested for those who add new scripts/integrations - and that's exactly the correct time to get those reports.

That being said, a warning is probably a safer starting point for those who take updates from upstream :)

Gnito: @OtterleyW so, after this discussion. Maybe we should add a warning to src/index.js if REACT_APP_CSP is not set. - And I'm suggesting that you add it to this PR.

OtterleyW: @Gnito Having the warning e.g. in frontend console sounds like a good idea. Do you think it should be included in this PR? And if yes, would adding the warning index.js make sense or is there some other place it would fit better?

src/index.js
```diff
@@ -106,6 +106,16 @@ if (typeof window !== 'undefined') {
   }
 }
 
+// Show warning if CSP is not enabled
+const CSP = process.env.REACT_APP_CSP;
+const cspEnabled = CSP === 'block' || CSP === 'report';
+
+if (!cspEnabled) {
```
Gnito, line 113: Another suggestion to add:
```js
if (CSP === 'report' && process.env.REACT_APP_ENV === 'production') {
  console.warn('Your production environment should use CSP with "block" mode. Read more form: https://www.sharetribe.com/docs/ftw-security/how-to-set-up-csp-for-ftw/');
} else if (!cspEnabled) { //...
```

Maybe I'm a bit too paranoid :)

src/index.js
```diff
@@ -106,6 +106,16 @@ if (typeof window !== 'undefined') {
...
+  console.warn(
+    "CSP is currently not enabled! You should add an environment variable REACT_APP_CSP with the value 'report' or 'block'. Read more form: https://www.sharetribe.com/docs/ftw-security/how-to-set-up-csp-for-ftw/"
```
Gnito, line 115:
```suggestion
    "CSP is currently not enabled! You should add an environment variable REACT_APP_CSP with the value 'report' or 'block'. Read more from: https://www.sharetribe.com/docs/ftw-security/how-to-set-up-csp-for-ftw/"
```

CHANGELOG.md
```diff
@@ -14,16 +14,18 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2020-XX-XX
 
+- [change] Update Helmet to v4.0.0 [#1347](https://github.com/sharetribe/ftw-daily/pull/1347)
```
Gnito, line 17: Also, add some comment about these new changes to here

PR #1348, opened 2020-08-13
Update path-to-regexp v6.1.0

Gnito:
path-to-regexp (v3.0.0 -> v6.1.0) **doesn't contain default expor**t anymore.


OtterleyW (approved):
👍

Issue #1338, opened 2020-08-10
JSON-LD Metadata is rendered duplicated - SEO


3wnbr1:
**Describe the bug** A clear and concise description of what the bug is.

Once the webpage is done rendering, the ``<script type="application/json+ld">...</script>``` templated through Helmet in Page component appears duplicated.

**To Reproduce** Steps to reproduce the behavior:

1. Get ftw-daily runinng in production mode (also works in dev mode but less obvious)
2. Open the devtools in your browser
3. Notice that ```<script type="application/json+ld">...</script>``` appears both at the bottom of the header and the bottom of the served page.

**Expected behavior** A clear and concise description of what you expected to happen.

```<script type="application/json+ld">...</script>``` should only appear once, this causes issues with Google Rich Results testing tools and furthermore, with SEO.

**Desktop (please complete the following information):**

- OS: macOS 10.15.6
- Browser : Safari : 13.x / 14.0 / Chrome 84.0.x

**Smartphone (please complete the following information):**

- Device: iPhone SE (1st gen)
- OS: iOS 14 Developper Beta 4
- Browser : Safari
- Version : 14

**Googlebot also encounters the issue !!!**

** Similar issues **

Even though ftw-daily uses `react-helmet-async`, there seems to be a bug in the way Hemet templates script tags

https://github.com/nfl/react-helmet/issues/486
https://github.com/nfl/react-helmet/issues/183#issuecomment-293047444

**Towards a fix***

Applying this patch mitigates the issue and therefore, json-ld metadata becomes valid for Google
```patch
diff --git a/src/components/Page/Page.js b/src/components/Page/Page.js
index 34243b6c..fd968404 100644
--- a/src/components/Page/Page.js
+++ b/src/components/Page/Page.js
@@ -43,6 +43,12 @@ class PageComponent extends Component {
     // handling both dragover and drop events.
     document.addEventListener('dragover', preventDefault);
     document.addEventListener('drop', preventDefault);
+
+    // Fix duplicate JSON-LD
+    const jsons = document.querySelectorAll('script[type="application/ld+json"]');
+    if (jsons.length > 0) {
+      jsons.item(0).remove();
+    }
   }
 
   componentWillUnmount() {
```

Gnito:
@3wnbr1 thanks for reporting - this is an interesting problem. 

Initial thoughts:

[...]

PR #1355, opened 2020-08-23
Fix: remove  douplicated page schema

Gnito:
This removes duplicate schema (LD+JSON) script element from the DOM. CSR removes #page-schema from DOM if one exists. We use "id" attribute to identify correct schema script (multiple schemas should be OK for search engines).

Note: Google schema debugger does notice that there was one script coming from SSR and CSR adds another on initial page load.


Gnito:
This is a fix to issue #1338

OtterleyW (approved):
LGTM 👍

PR #1361, opened 2020-09-18
Improve error handling by passing error details forward 

OtterleyW:
Improve error handling by passing error details forward instead of creating a new error that hides the details.

Using `e.apiResponse` directly didn't pass the data forward. Because the error details are saved to response.json and response.json() returns promise, we need to wait for that promise to resolve before we can access the data. 

Before:
Screenshot 2020-09-18 at 14 17 13

After:
Screenshot 2020-09-18 at 14 16 38

### DIFF
--- a/src/util/api.js
+++ b/src/util/api.js
@@ -46,25 +46,25 @@ const post = (path, body) => {
     },
     body: serialize(body),
   };
-  return window
-    .fetch(url, options)
-    .then(res => {
-      const contentTypeHeader = res.headers.get('Content-Type');
-      const contentType = contentTypeHeader ? contentTypeHeader.split(';')[0] : null;
-      if (contentType === 'application/transit+json') {
-        return res.text().then(deserialize);
-      } else if (contentType === 'application/json') {
-        return res.json();
-      }
-      return res.text();
-    })
-    .then(res => {
-      if (res.status >= 400) {
-        const e = res;
+  return window.fetch(url, options).then(res => {
+    const contentTypeHeader = res.headers.get('Content-Type');
+    const contentType = contentTypeHeader ? contentTypeHeader.split(';')[0] : null;
+
+    if (res.status >= 400) {
+      return res.json().then(data => {
+        let e = new Error();
+        e = Object.assign(e, data);
+
         throw e;
-      }
-      return res;
-    });
+      });
+    }
+    if (contentType === 'application/transit+json') {
+      return res.text().then(deserialize);
+    } else if (contentType === 'application/json') {
+      return res.json();
+    }
+    return res.text();
+  });
 };
 
 // Fetch transaction line items from the local API endpoint.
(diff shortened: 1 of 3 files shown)


src/util/api.js
```diff
@@ -65,6 +57,13 @@ const post = (path, body) => {
         return res.json();
       }
       return res.text();
+    })
+    .then(res => {
```
Gnito, line 61: If I understand this correctly, this is not anymore "Response" object but resolved data (either res.text() or res.json()).

I'd suggest that you change the param name to `data`.

src/util/api.js
```diff
@@ -65,6 +57,13 @@ const post = (path, body) => {
...
+      if (res.status >= 400) {
+        const e = res;
+        throw e;
+      }
+      return res;
```
kpuputti, line 66: Is `res` an instance of Error here? I think it's a generic convention in JavaScript to only throw Error instances, even if the language allows throwing pretty much anything.

I guess this is just a convenient way to trigger the default error printing, right?

src/util/api.js
```diff
@@ -65,6 +57,13 @@ const post = (path, body) => {
         return res.json();
       }
       return res.text();
+    })
+    .then(res => {
+      if (res.status >= 400) {
```
Gnito, line 62: What happens if text() is used? I.e. 
- `res.text().then(data => data.status)` or 
- `res.text().then(deserialize).then(data => data.status)`

I assume that the status is in those cases "undefined". 

I kind of feel that this should be handled differently. Maybe you could follow the "serialize" pattern if the error is only possible with 'application/json' format type?

src/util/api.js
Gnito: @kpuputti I guess the problem is that the "res" is a JSON object. Using plain Error is a bit problematic since its constructor is quite picky compared to plain _throw_. So, it's either `throw data;` or `throw new MyCustomError(data);`

If I read this correctly, Axios uses plain Promise.reject (i.e. pretty much same approach as plain "throw"): https://github.com/axios/axios/blob/0d69a79c81a475f1cca6d83d824eed1e5b0b045d/lib/core/dispatchRequest.js#L77

OtterleyW: @kpuputti Good point! I didn't even think about this because it was possible to throw the data object directly. I will change this to Error instance,

src/util/api.js
OtterleyW: @Gnito good catch! I I'm not sure how this should be handled. What do you mean by following the "serialize" pattern?

Gnito: Sorry, I meant to write "deserialize" to refer to existing function. I.e. I was referring to having a completely separate function that handles error printing (similarly as `res.text().then(deserialize)`). 

Anyway, there could be a custom path for error handling when `if (res.status >= 400) {` the "throw" is anyway breaking the "then" chain.

Gnito: I think you could write something like:

```js
if (res.status >= 400) {
  return res.json().then(throwErrorHelper);
} 
```

OtterleyW: @Gnito That makes sense, thanks for explaining! I changed the error handling so do you think it's ok now?

PR #1364, opened 2020-09-22
Add Facebook login to FTW

OtterleyW:
This is a base branch for adding a Facebook login. On the high level this PR contains the following changes:

*Generic:*
- New env variables `REACT_APP_FACEBOOK_APP_ID` and `FACEBOOK_APP_SECRET`. The Facebook app id must be exposed to the frontend (using prefix REACT_APP_) so that we can determine if the Facebook login should be enabled or not. 
- New cookies `st-authinfo` and `st-autherror` are used to store information. The expiration time for both cookies is 15 minutes.

*Server:*
- New endpoints `createUserWithIdp` and `loginWithIdp` - for using these, you need the latest version of `sharetribe-flex-sdk`
- Add [Passport.js](http://www.passportjs.org/) library for authenticating 
- Use Passport's Facebook strategy to authenticate with Facebook. New routes `auth/facebook` and `auth/facebook/callback`

*AuthenticationPage:*
- Add button for logging in or signing up with Facebook and new function `authWithFacebook`
- Add new `ConfirmForm` form confirming information fetched from the identity provider (IdP, e.g. Facebook)
- Read information about loginWithIdp call from cookies

*Button:*
- Add new variant for the social login button

*CheckoutPage and ListingPage:*
- Add new parameter `saveToSessionStorage` to `setInitialValues` function so that we can store initial values to session storage before user is redirected to Facebook login. This enables redirecting user directly to CheckoutPage after authenticating from ListingPage

*ContactDetailsPage and PasswordChangePage:*
- Add reset password link to the pages where the current password is required (e.g. updating email). When signing up with IdP (e.g. Facebook), the user doesn't have an existing password so they first need to create one by going through the reset password flow.

---

This can be merged after the following parts have been completed:
- [x] `authWithIdp` and `createWithIdp` endpoints to server (#1365)
- [x]  Facebook login (#1366)
- [x] Add password reset links (#1367)


src/ducks/Auth.duck.js
```diff
@@ -48,6 +53,10 @@ const initialState = {
   // signup
   signupError: null,
   signupInProgress: false,
+
+  // confirm (create use with idp)
+  confirmErro: null,
```
feorgie, line 58: can I confirm, Is this a typo?

OtterleyW: @feorgie yes that's a typo, it should be `confirmError`. Thanks for letting us know!

feorgie: no problem!

PR #1365, opened 2020-09-22
Add idp endpoints to FTW server

OtterleyW:
This PR adds `authWithIdp` and `createWithIdp` endpoints to FTW server which enables authenticating and signing up to Flex using access token from 3rd party identity provider (e.g. Facebook or Google)

When done, this branch will be merged to social-logins base branch.

### DIFF
--- a/server/apiRouter.js
+++ b/server/apiRouter.js
@@ -16,6 +16,8 @@ const transactionLineItems = require('./api/transaction-line-items');
 const initiatePrivileged = require('./api/initiate-privileged');
 const transitionPrivileged = require('./api/transition-privileged');
 
+const createUserWithIdp = require('./api/auth/createUserWithIdp');
+
 const router = express.Router();
 
 // ================ API router middleware: ================ //
@@ -49,5 +51,7 @@ router.get('/login-as', loginAs);
 router.post('/transaction-line-items', transactionLineItems);
 router.post('/initiate-privileged', initiatePrivileged);
 router.post('/transition-privileged', transitionPrivileged);
+
+router.post('/auth/create-user-with-idp', createUserWithIdp);
 
 module.exports = router;
--- a/src/util/api.js
+++ b/src/util/api.js
@@ -1,3 +1,7 @@
+// These helpers are calling FTW's own server-side routes
+// and nor directly calling Marketplace API or Integration API
+// You can find these api endpoints from 'server/api/...' directory
+
 import { types as sdkTypes, transit } from './sdkLoader';
 import config from '../config';
 import Decimal from 'decimal.js';
@@ -97,4 +101,17 @@ export const initiatePrivileged = body => {
 // be sent in the body.
 export const transitionPrivileged = body => {
   return post('/api/transition-privileged', body);
+};
+
+// Create user with identity provider (e.g. Facebook or Google)
+//
+// If loginWithIdp api call fails and user can't authenticate to Flex with idp
+// we will show option to create a new user with idp.
+// For that user needs to confirm data fetched from the idp.
+// After the confrimation, this endpoint is called to create a new user with confirmed data.
+//
+// See `server/api/auth/createUserWithIdp.js` to see what data should
+// be sent in the body.
+export const createUserWithIdp = body => {
+  return post('/api/auth/create-user-with-idp', body);
 };
--- a/server/api/auth/createUserWithIdp.js
+++ b/server/api/auth/createUserWithIdp.js
@@ -1,0 +1,81 @@
+const http = require('http');
+const https = require('https');
+const sharetribeSdk = require('sharetribe-flex-sdk');
+const { handleError, serialize, typeHandlers } = require('../../api-util/sdk');
+
+const CLIENT_ID = process.env.REACT_APP_SHARETRIBE_SDK_CLIENT_ID;
+const CLIENT_SECRET = process.env.SHARETRIBE_SDK_CLIENT_SECRET;
+const TRANSIT_VERBOSE = process.env.REACT_APP_SHARETRIBE_SDK_TRANSIT_VERBOSE === 'true';
+const USING_SSL = process.env.REACT_APP_SHARETRIBE_USING_SSL === 'true';
+const BASE_URL = process.env.REACT_APP_SHARETRIBE_SDK_BASE_URL;
+
+const FACBOOK_APP_ID = process.env.REACT_APP_FACEBOOK_APP_ID;
+const GOOGLE_CLIENT_ID = process.env.REACT_APP_GOOGLE_CLIENT_ID;
+
+const FACEBOOK_IDP_ID = 'facebook';
+const GOOGLE_IDP_ID = 'google';
+
+// Instantiate HTTP(S) Agents with keepAlive set to true.
+// This will reduce the request time for consecutive requests by
+// reusing the existing TCP connection, thus eliminating the time used
+// for setting up new TCP connections.
+const httpAgent = new http.Agent({ keepAlive: true });
+const httpsAgent = new https.Agent({ keepAlive: true });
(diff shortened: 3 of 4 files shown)


src/util/api.js
```diff
@@ -98,3 +98,15 @@ export const initiatePrivileged = body => {
...
+//
+// If the authentication with idp fails, we will create a new user.
+// For that user needs to confirm data fetched from the idp.
+// This endpoint creates the new user with confirmed data.
+//
+// See `server/api/auth/createWithIdp.js` to see what data should
+// be sent in the body.
+export const createWithIdp = body => {
+  return post('/api/auth/create-with-idp', body);
+};
```
Gnito, line 112: Sorry, this is not really related to this PR, but I feel that this util/api.js is pretty confusing file due to naming. 

Could we add some description to the beginning of the file (or somewhere) that these helpers are about client app's server-side routes (not something that is directly calling Marketplace API and Integration API)? 

When we mention API in Flex Docs / Slack / Intercom, it's pretty much always about those Flex APIs instead of these client app's own server routes (which currently are a kind of "proxy" calls, but that's likely to change). So, this file might cause some confusion without better explanation somewhere.

server/apiRouter.js
```diff
@@ -50,4 +52,6 @@ router.post('/transaction-line-items', transactionLineItems);
 router.post('/initiate-privileged', initiatePrivileged);
 router.post('/transition-privileged', transitionPrivileged);
 
+router.post('/auth/create-with-idp', createWithIdp);
```
Gnito, line 55: I find this server route quite confusing. What are we creating? Could the name be more self-explanatory: `create-user-with-idp`?

(and then related variable names should be changed accordingly)

Gnito: I.e. `/auth/auth-with-idp` is kind of understandable, but `/auth/create-with-idp` makes you wonder what's happening.

server/apiRouter.js
```diff
@@ -16,6 +16,8 @@ const transactionLineItems = require('./api/transaction-line-items');
 const initiatePrivileged = require('./api/initiate-privileged');
 const transitionPrivileged = require('./api/transition-privileged');
 
+const createWithIdp = require('./api/auth/createWithIdp');
```
Gnito, line 19: Where's `authWithIdp`?

server/api/auth/createWithIdp.js
```diff
@@ -0,0 +1,82 @@
...
+    typeHandlers,
+    ...baseUrl,
+  });
+
+  const { idpToken, source, ...rest } = req.body;
+
+  // Choose the idpClientId based on which authentication method is used.
+  // Currently available options are "Facebook" and "Google"
+  const idpClientId =
+    source === 'Facebook' ? FACBOOK_APP_ID : source === 'Google' ? GOOGLE_CLIENT_ID : null;
```
Gnito, line 48: Maybe these source options should be added as constants somewhere (in this file or some other place). Then there are fewer chances for typos in the future customizations since unknown variables cause errors.

server/api/auth/authWithIdp.js
```diff
@@ -0,0 +1,97 @@
...
+    .authWithIdp({
+      idpClientId: clientID,
+      idpToken: user ? user.accessToken : null,
+    })
+    .then(response => {
+      if (response.status === 200) {
+        // If the user was authenticated, redirect back to to LandingPage
+        // We need to add # to the end of the URL because otherwise Facebook
+        // login will add their defaul #_#_ which breaks the routing in frontend.
+        res.redirect(`${rootUrl}/#`);
```
Gnito, line 71: Will this break our routing in a scenario where anonymous user clicks "book" button on listing page? 

I.e. Frontend should 
1) redirect an unauthenticated user to our authentication page, 
2) then user authenticates
3) and finally, user is automatically redirected to checkout page after successful login

server/api/auth/authWithIdp.js
```diff
@@ -0,0 +1,97 @@
...
+    })
+    .catch(() => {
+      // If authentication fails, we want to create a new user with idp
+      // For this we will need to pass some information to frontend so
+      // that we can use that information in createWithIdp api call.
+      // The createWithIdp api call is triggered from frontend
+      // after showing a confirm page to user
+
+      res.cookie(
+        'authinfo',
```
Gnito, line 82: We should document our cookies somewhere. Previously, we had just 2 cookies of our own (sdk & cookie-consent cookie). Then integrations brought their own cookies: Stripe, map provider, etc.

This specific info should be cleared from cookies (at least, on logout)
https://github.com/sharetribe/ftw-daily/blob/master/src/reducers.js

server/apiRouter.js
Gnito: Or is it omitted on purpose?

server/api/auth/createWithIdp.js
```diff
@@ -0,0 +1,82 @@
...
+          serialize({
+            status,
+            statusText,
+            data,
+          })
+        )
+        .end();
+    })
+    .catch(e => {
+      handleError(res, e);
```
Gnito, line 80: What happens to cookie?
(I assume that it should be cleared here (and ensure it also on logout)

server/api/auth/authWithIdp.js
```diff
@@ -0,0 +1,97 @@
...
+        {
+          status: err.status,
+          code: err.code,
+          message: err.message,
+        },
+        {
+          maxAge: 15 * 60 * 1000, // 15 minutes
+        }
+      )
+      .redirect(`${rootUrl}/login#`);
```
Gnito, line 40: This is assuming that the web app has a route called "login" and customizers haven't changed the route configuration.

server/apiRouter.js
lyyder: The corresponding Flex API endpoint is `/current_user/create_with_idp`. There the `/current_user` path gives context to the "create" part. So how about `/auth/create-user-with-idp`?

src/util/api.js
```diff
@@ -98,3 +98,15 @@ export const initiatePrivileged = body => {
 export const transitionPrivileged = body => {
   return post('/api/transition-privileged', body);
 };
+
+// Create user with identity provider (e.g. Facebook or Google)
+//
+// If the authentication with idp fails, we will create a new user.
+// For that user needs to confirm data fetched from the idp.
+// This endpoint creates the new user with confirmed data.
```
lyyder, line 106: Is this correct? So if the user can not authenticate to th IdP, we'll create a user? Or if the authentication succeeds, we'll create a user with confirmed data from the IdP.

server/apiRouter.js
Gnito: @lyyder yeah, that's what I was after in the first comment. I just forgot the "/auth/" part.

server/apiRouter.js
OtterleyW: We are using `authWithIdp` only internally in FTW server (e.g. here: https://github.com/sharetribe/ftw-daily/pull/1366/files#diff-c2569a98b269447cc716700472d37356R51) so there is no specific need for adding a route for it. 

First I had the authWithIdp under `api-util` instead of `api/auth` but that felt also confusing.

server/api/auth/createWithIdp.js
OtterleyW: The cookie will be cleared if the authentication is successful or it will expire after 15 min (or this time can be changed). 

I'm not sure if the cookie should be cleared here when there is an error because then the user is not able to try the confirm step again. E.g. the error could be a conflicting email and then the user could try again with a different email (but I'm not sure how likely something like this would happen. One scenario would be that user has signed up with the same email as in e.g. Facebook but not verified the email address yet and for some reason they would decide to use different email in Flex).

PR #1366, opened 2020-09-22
Add Facebook login

OtterleyW:
**Adds Facebook login to FTW**

This is how the authentication flow works in general:

Screenshot 2020-10-02 at 13 40 38
Screenshot 2020-10-02 at 13 41 22

(these graphs are also in Whimsical Socia logins & SSO board https://whimsical.com/GcZr7nhZwex4WfYgAuiHFG)

Screenshot 2020-09-22 at 14 37 02

If a new user needs to be created there is a new `ConfirmSignupForm` on `ConfirmPage` to confirm the data fetched from e.g. Facebook:
Screenshot 2020-09-22 at 14 36 42

This PR can be merged after #1365 is merged to the base branch #1364


src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -54,14 +72,22 @@ export class AuthenticationPageComponent extends Component {
...
       submitLogin,
       submitSignup,
+      confirmError,
+      submitSinguoWithIdp,
       tab,
       sendVerificationEmailInProgress,
       sendVerificationEmailError,
       onResendVerificationEmail,
       onManageDisableScrolling,
+      showSocialLoginsForTests,
```
OtterleyW, line 82: The snapshot test for AuthenticationPage caused some issues because in the local environment facebook app id was in environment variables and in CI it was not. Because of this, the social login button was added to snapshot when running tests locally but in CI the page was expected to be rendered without the social login button. 

Adding extra prop feels really hacky but I'm not sure in what another way we could ensure that the page is rendered similarly locally and in CI. Also with the prop, it was possible to test both options. 

Do you have any better ideas on how to handle the snapshot test (or do we even need to keep it)?

src/containers/ListingPage/ListingPage.js
```diff
@@ -59,6 +59,10 @@ import SectionRulesMaybe from './SectionRulesMaybe';
 import SectionMapMaybe from './SectionMapMaybe';
 import css from './ListingPage.css';
 
+import { storeData } from '../CheckoutPage/CheckoutPageSessionHelpers';
+// This storage key is used in sessionStorage and it needs to be same as on CheckoutPage
+const STORAGE_KEY = 'CheckoutPage';
+
```
OtterleyW, line 65: We need to call `storeData` function already on ListingPage so that the user can be redirected directly to CheckoutPage after authenticating with idp (Facebook). 

Maybe we should move the session helpers file from CheckoutPage to utils and change the storage key to something else now that this is not restricted just to CheckoutPage. But I wanted to know if this setup makes sense before doing this change.

Gnito: @OtterleyW Why can't you use "callSetInitialValues(setInitialValues, initialValues);"?

src/util/api.js
```diff
@@ -1,8 +1,12 @@
+// These helpers are calling FTW's own server-side routes
+// and nor directly calling Marketplace API or Integration API
```
Gnito, line 2:
```suggestion
// So, they are not directly calling Marketplace API or Integration API.
```

src/util/api.js
```diff
@@ -98,3 +102,16 @@ export const initiatePrivileged = body => {
 export const transitionPrivileged = body => {
   return post('/api/transition-privileged', body);
 };
+
+// Create user with identity provider (e.g. Facebook or Google)
+//
+// If loginWithIdp api call fails and user can't authenticate to Flex with idp
+// we will show option to create a new user with idp.
+// For that user needs to confirm data fetched from the idp.
+// After the confrimation, this endpoint is called to create a new user with confirmed data.
```
Gnito, line 111:
```suggestion
// After the confirmation, this endpoint is called to create a new user with confirmed data.
```

server/apiRouter.js
```diff
@@ -50,4 +54,9 @@ router.post('/transaction-line-items', transactionLineItems);
 router.post('/initiate-privileged', initiatePrivileged);
 router.post('/transition-privileged', transitionPrivileged);
 
+router.post('/auth/create-user-with-idp', createUserWithIdp);
+
+router.get('/auth/facebook', authenticateFacebook);
```
Gnito, line 59: These might benefit from some extra comment.

server/index.js
```diff
@@ -145,6 +146,9 @@ if (!dev) {
   }
 }
 
+// Initialize passport
```
Gnito, line 149: What is passport? 
(I.e. some extra comment would make this more meaningful.)

src/config.js
```diff
@@ -101,11 +101,14 @@ const siteInstagramPage = null;
...
 
+// Social logins & SSO
+
+// Note: Facebook app id is also used for tracking:
 // Facebook counts shares with app or page associated by this id
 // Currently it is unset, but you can read more about fb:app_id from
 // https://developers.facebook.com/docs/sharing/webmasters#basic
 // You should create one to track social sharing in Facebook
-const facebookAppId = null;
+const facebookAppId = process.env.REACT_APP_FACEBOOK_APP_ID;
```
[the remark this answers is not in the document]
Gnito: Me neither. (So, your decision sounds good to me.)

src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -136,10 +172,88 @@ export class AuthenticationPageComponent extends Component {
...
+        idpId,
+        ...authParams,
+      });
+    };
+
+    const authWithFacebook = () => {
+      const baseUrl = apiBaseUrl();
+
+      if (from) {
+        window.location.href = `${baseUrl}/api/auth/facebook?from=${from}`;
```
Gnito, line 199: I really think we should include other possible return paths as URL params too!

That way we don't need to hard-code web app's internal routes all over the place.
I think **landing page** and **confirm page** are possible paths where the user might end up.

It would also be good to add some code comment to here. What is this "from"? what is it doing?

src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -136,10 +172,88 @@ export class AuthenticationPageComponent extends Component {
...
+      <div className={css.idpButtons}>
+        <SocialLoginButton onClick={() => authWithFacebook()}>
+          <span className={css.buttonIcon}>
+            <svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+              <path
+                d="M7.89.214C4.055 1.047 1.005 4.13.205 7.947c-.734 3.45.533 7.283 3.166 9.6.967.85 3.2 2.033 4.15 2.183l.617.1v-6.883H5.806v-3h2.283l.083-1.633c.134-2.417.717-3.534 2.3-4.25.617-.284 1.034-.35 2.3-.334.85.017 1.617.084 1.7.134.1.05.167.7.167 1.433v1.317h-.983c-1.484 0-1.75.283-1.817 1.983l-.067 1.35h1.45c1.284 0 1.434.033 1.35.283-.05.167-.133.667-.2 1.134-.216 1.55-.25 1.583-1.483 1.583h-1.083V19.914l.866-.234c1.684-.433 2.984-1.216 4.4-2.633 2.067-2.067 2.9-4.1 2.9-7.017 0-3.166-1.2-5.75-3.616-7.766C14.106.38 10.772-.42 7.889.214z"
+                fill="#1877F2"
+                fillRule="nonzero"
+              />
+            </svg>
```
Gnito, line 223: It's probably better to make SVGs always as separate subcomponents (it could be inside this directory). 
The reasoning has been that people are not familiar with SVG syntax - having them in a separate component makes bug hunting a bit more straightforward (and people are not so tempted to tamper this unfamiliar code/colors/attributes).

Gnito: It's also easier to add comments like: 
"// This is Facebook's logo, you are not allowed to change its color"

src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -136,10 +172,88 @@ export class AuthenticationPageComponent extends Component {
...
+          inProgress={authInProgress}
+          onOpenTermsOfService={() => this.setState({ tosModalOpen: true })}
+          authInfo={this.state.authInfo}
+        />
+      </div>
+    ) : (
       <div className={css.content}>
         <LinkTabNavHorizontal className={css.tabs} tabs={tabs} />
         {loginOrSignupError}
+        {socialLoginButtonsMaybe}
```
Gnito, line 255: Is there some margin between loginOrSignupError and socialLoginButtonsMaybe?

src/containers/AuthenticationPage/AuthenticationPage.js
Gnito: Yeah, I agree, this feels a bit hacky.

Jest should automatically define NODE_ENV to be 'test'.
So, try adding the Facebook app id to .env.test (maybe as empty)

src/containers/AuthenticationPage/AuthenticationPage.js
Gnito: You could also lose the if-else pattern if there more params. You could create something like this:

```js
const fromParam = from ? `from=${from}` : '';
const defaultReturnParam = defaultReturn ? `&defaultReturn=${defaultReturn}` : '';
window.location.href = `${baseUrl}/api/auth/facebook?${fromParam}${defaultReturnParam}`;
```

BUT we do have [query-string](https://www.npmjs.com/package/query-string#stringifyurlobject-options) package as a dependency.

src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -136,10 +172,88 @@ export class AuthenticationPageComponent extends Component {
...
+          <FormattedMessage id="AuthenticationPage.loginWithFacebook" />
+        </SocialLoginButton>
+        <center>
+          <FormattedMessage id="AuthenticationPage.or" />
+        </center>
+      </div>
+    ) : null;
+
+    const formContent = isConfirm ? (
+      <div className={css.content}>
```
Gnito, line 234: One thing to consider in this kind of ternary-chain is that should one or more of these be subcomponent (inside this directory) to improve readability. 

(And I'm not saying that this is necessarily hard to read, but in Github, it's a bit confusing due to partially visible content.)

server/api/auth/loginWithIdp.js
```diff
@@ -0,0 +1,124 @@
...
+        'st-autherror',
+        {
+          status: err.status,
+          code: err.code,
+          message: err.message,
+        },
+        {
+          maxAge: 15 * 60 * 1000, // 15 minutes
+        }
+      )
```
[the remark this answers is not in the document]
Gnito: I'd vote Flex Docs.

server/api/auth/facebook.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+// ClientId is required when adding a new Facebook strategy to passport
+if (clientID) {
+  passport.use(new FacebookStrategy(strategyOptions, verifyCallback));
+}
+
+exports.authenticateFacebook = (req, res, next) => {
+  const from = req && req.query ? req.query.from : null;
+  passport.authenticate('facebook', { scope: ['email'], state: from })(req, res, next);
```
Gnito, line 52: How fixed this syntax is? 
I really think we should include other possible return paths as URL params too - and this "state" term is confusing. 

So, could this be something like:
```js
// pseudocode
const query = req && req.query;
const returnURL =  query && req.query.from 
  ? query.from 
  : query.defaultURL
  ? query.defaultURL 
  : null;
const confirmURL =  query && query.confirmURL ? query.confirmURL : null; 
passport.authenticate('facebook', { scope: ['email'], returnURL, confirmURL })(req, res, next);
```

Side note: I'm a bit confused about how these `from`/`state`/`user.returnUrl` work at the moment. Maybe some parameter renaming would help?

Gnito (changes requested):
I added some comments. 
I'm not sure I'm completely aware of all the aspects, but I still think we should parameterize all the possible return paths.

src/translations/en.json
```diff
@@ -17,11 +17,16 @@
   "ActivityFeed.transitionExpire": "Booking request expired. { displayName } did not reply on time.",
   "ActivityFeed.transitionRequest": "{ displayName } requested to book { listingTitle }.",
   "ActivityFeed.transitionReview": "{ displayName } left a review for you. { reviewLink }",
+  "AuthenticationPage.confirmSignupInfoText": "Please check that your information is correct.",
+  "AuthenticationPage.confirmSignupWithIdpTitle": "Signup with {idp}",
```
Gnito, line 21: One more comment: This idp variable was in lower case: "Signup with facebook".
So, before you pass the variable to FormattedMessage, you should capitalize it.

PR #1367, opened 2020-10-12
Add password reset link

OtterleyW:


src/containers/ContactDetailsPage/ContactDetailsPage.duck.js
```diff
@@ -12,13 +12,19 @@ export const SAVE_PHONE_NUMBER_ERROR = 'app/ContactDetailsPage/SAVE_PHONE_NUMBER
 
 export const SAVE_CONTACT_DETAILS_CLEAR = 'app/ContactDetailsPage/SAVE_CONTACT_DETAILS_CLEAR';
 
+export const RESET_PASSWORD_REQUEST = 'app/PasswordResetPage/RESET_PASSWORD_REQUEST';
+export const RESET_PASSWORD_SUCCESS = 'app/PasswordResetPage/RESET_PASSWORD_SUCCESS';
+export const RESET_PASSWORD_ERROR = 'app/PasswordResetPage/RESET_PASSWORD_ERROR';
```
lyyder, line 17: Would it make sense to import these from PasswordResetPage duck? Do we have a pattern for using other page's states in another duck?

Gnito: I assume this is a copy-paste typo. Rename this action strings correctly or our page-reducers start to react other pages changes. (One option is to create public duck-file, but that might be an overkill.

OtterleyW: Yeah, that's a copy-paste typo (it seems I never remember to change them...) I'll fix that!

I also considered a shared duck file for these but it really felt a bit overkill for now.

PR #1376, opened 2020-10-29
Add Google login to FTW

OtterleyW:
This PR adds Google login to FTW

It follows the same logic as Facebook login and uses the `createUserWithIdp` and `loginWithIdp` endpoints which were added with Facebook login. You can find more information from Facebook login PR #1366


server/api/auth/loginWithIdp.js
```diff
@@ -99,7 +99,9 @@ module.exports = (err, user, req, res, clientID, idpId) => {
         }
       }
     })
-    .catch(() => {
+    .catch(err => {
+      console.error(err);
```
Gnito, line 103: What kind of data the 'err' contains? Is there something that should not end up to logs?

Gnito: Maybe all the console.logs in the server routes could be checked and potentially enforce the usage of [log.js](https://github.com/sharetribe/ftw-daily/blob/master/server/log.js) instead?

src/forms/ConfirmSignupForm/ConfirmSignupForm.js
```diff
@@ -25,6 +25,7 @@ const ConfirmSignupFormComponent = props => (
         intl,
         onOpenTermsOfService,
         authInfo,
+        idp,
       } = fieldRenderProps;
```
Gnito, line 29: Btw. I just noticed that an old typo keeps spreading around. These are not fieldRenderProps but formRenderProps:
https://final-form.org/docs/react-final-form/types/FormRenderProps

It's a bit annoying because it leads to the wrong place when someone tries to debug forms against React Final Form's docs.

(Side note: I think this typo might even be copied Final Form's docs originally.)

server/api/auth/google.js
```diff
@@ -0,0 +1,87 @@
+const passport = require('passport');
+const GoogleStrategy = require('passport-google-oauth').OAuth2Strategy;
+const loginWithIdp = require('./loginWithIdp');
+
+const radix = 10;
+const PORT = parseInt(process.env.REACT_APP_DEV_API_SERVER_PORT, radix);
+const rootUrl = process.env.REACT_APP_CANONICAL_ROOT_URL;
+const clientID = process.env.REACT_APP_GOOGLE_CLIENT_ID;
+const clientSecret = process.env.GOOGLE_CLIENT_SECRET;
```
lyyder, line 9: Wondering what this is used for. The Google Web login reference does not mention using the client secret: https://developers.google.com/identity/sign-in/web/reference. Apparently this is [used to fetch user profile info](https://github.com/jaredhanson/passport-oauth2/blob/fda4a53fe63ac13b825aec0ae87a4d6bb599ebd5/lib/strategy.js#L95) but shouldn't the access token suffice for that.

What if this is left out? Does fetching the profile fail?

lyyder (commented):
Looks good 👍. Only suggestion is to try the login once without the client secret and see if a) the login succeeds b) fetching user profile succeeds.

server/api/auth/google.js
OtterleyW: I tested this without clientSecret and it didn't work. At least the Passport.js expects that the client secret is there. After some investigation I think I found the answer from [Google docs](https://developers.google.com/identity/protocols/oauth2#installed): For example, a JavaScript application does not require a secret, but a web server application does.

It's a little confusing because we are also using JavaScript in the FTWs server. But anyway, when we don't use Google's JavaScript client directly for the web application but we are handling the authentication in the web server, we need to provide the client secret too.

PR #1379, opened 2020-11-03
Fix 401 rendering on SSR. Prev bug: token could be null too

Gnito:
Previous code was broken since token could have been null.
I.e. something like this could have worked too: `!token || (!!scopes && scopes.length === 1 && scopes[0] === 'public-read');`


lyyder (approved):
LGTM 👍

PR #1380, opened 2020-11-03
Fix double rendering of Routes.js

Gnito:
Redux container (connect HOC) caused double rendering for Routes component. 
Refactored the code to keep the React Router setup less error-prone and to avoid double navigation, which also manifested as calling the loadData function twice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2020-XX-XX
 
+- [fix] Routes component got double rendered due to Redux container HOC. Because navigation could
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2020-XX-XX
 
+- [fix] Routes component got double rendered due to Redux container HOC.
+  [#1380](https://github.com/sharetribe/ftw-daily/pull/1380)
```
kpuputti, line 18: You could mention here that this caused `loadData` to be called twice in client-side page changes.

kpuputti (approved):
I didn't test this, but looks good!

PR #1383, opened 2020-12-02
OIDC proxy implementation

OtterleyW:
In this PR, we add helper functions for using FTW as OIDC proxy.  These new helpers are added in the `api-util/idToken.js` file. This also introduces two now server-side endpoints for serving _discovery document_ and _jwks uri_.

### DIFF
--- a/server/api-util/idToken.js
+++ b/server/api-util/idToken.js
@@ -1,0 +1,100 @@
+const crypto = require('crypto');
+const { default: fromKeyLike } = require('jose/jwk/from_key_like');
+const { default: SignJWT } = require('jose/jwt/sign');
+
+const radix = 10;
+const PORT = parseInt(process.env.REACT_APP_DEV_API_SERVER_PORT, radix);
+const rootUrl = process.env.REACT_APP_CANONICAL_ROOT_URL;
+const useDevApiServer = process.env.NODE_ENV === 'development' && !!PORT;
+
+const issuerUrl = useDevApiServer ? `http://localhost:${PORT}/api` : `${rootUrl}/api`;
+
+/**
+ * Gets user information and creates the signed jwt for id token.
+ *
+ * @param {string} idpClientId the client id of the idp provider in Console
+ * @param {Object} options signing options containing signingAlg and required key information
+ * @param {Object} user user information containing at least firstName, lastName, email and emailVerified
+ *
+ * @return {Promise} idToken
+ */
+exports.createIdToken = (idpClientId, user, options) => {
+  if (!idpClientId) {
+    console.error('Missing idp client id!');
+    return;
+  }
+  if (!user) {
+    console.error('Missing user information!');
+    return;
+  }
+
+  const signingAlg = options.signingAlg;
+
+  // Currently Flex supports only RS256 signing algorithm.
+  if (signingAlg !== 'RS256') {
+    console.error(`${signingAlg} is not currently supported!`);
+    return;
+  }
+
+  const { rsaPrivateKey, keyId } = options;
+
+  if (!rsaPrivateKey) {
+    console.error('Missing RSA private key!');
+    return;
+  }
+
+  // We use jose library which requires the RSA key
+  // to be KeyLike format:
+  // https://github.com/panva/jose/blob/master/docs/types/_types_d_.keylike.md
+  const privateKey = crypto.createPrivateKey(rsaPrivateKey);
+
+  const { userId, firstName, lastName, email, emailVerified } = user;
+
+  const jwt = new SignJWT({
+    given_name: firstName,
+    family_name: lastName,
+    email: email,
+    email_verified: emailVerified,
+  })
+    .setProtectedHeader({ alg: signingAlg, kid: keyId })
+    .setIssuedAt()
+    .setIssuer(issuerUrl)
+    .setSubject(userId)
+    .setAudience(idpClientId)
+    .setExpirationTime('1h')
+    .sign(privateKey);
+
+  return jwt;
+};
+
+// Serves the discovery document in json format
+// this document is expected to be found from
+// api/.well-known/openid-configuration endpoint
+exports.openIdConfiguration = (req, res) => {
+  res.json({
+    issuer: issuerUrl,
+    jwks_uri: `${issuerUrl}/.well-known/jwks.json`,
+    subject_types_supported: ['public'],
+    id_token_signing_alg_values_supported: ['RS256'],
+  });
+};
+
+/**
+ * @param {String} signingAlg signing algorithm, currently only RS256 is supported
+ * @param {Array} list containing keys to be served in json endpoint
+ *
+ * // Serves the RSA public key as JWK
+// this document is expected to be found from
+// api/.well-known/jwks.json endpoint as stated in discovery document
+ */
+exports.jwksUri = keys => (req, res) => {
+  const jwkKeys = keys.map(key => {
+    return fromKeyLike(crypto.createPublicKey(key.rsaPublicKey)).then(res => {
+      return { alg: key.alg, kid: key.keyId, ...res };
+    });
+  });
+
+  Promise.all(jwkKeys).then(resolvedJwkKeys => {
+    res.json({ keys: resolvedJwkKeys });
+  });
+};
--- a/server/api/auth/loginWithIdp.js
+++ b/server/api/auth/loginWithIdp.js
@@ -20,7 +20,7 @@ const httpsAgent = new https.Agent({ keepAlive: true });
 
 const baseUrl = BASE_URL ? { baseUrl: BASE_URL } : {};
 
-module.exports = (err, user, req, res, clientID, idpId) => {
+module.exports = (err, user, req, res, idpClientId, idpId) => {
   if (err) {
     log.error(err, 'fetching-user-data-from-idp-failed');
 
@@ -87,7 +87,7 @@ module.exports = (err, user, req, res, clientID, idpId) => {
   return sdk
     .loginWithIdp({
       idpId,
-      idpClientId: clientID,
+      idpClientId,
       idpToken: user.idpToken,
     })
     .then(response => {
--- a/server/apiRouter.js
+++ b/server/apiRouter.js
@@ -21,6 +21,8 @@ const createUserWithIdp = require('./api/auth/createUserWithIdp');
 const { authenticateFacebook, authenticateFacebookCallback } = require('./api/auth/facebook');
 const { authenticateGoogle, authenticateGoogleCallback } = require('./api/auth/google');
 
+const { openIdConfiguration, jwksUri } = require('./api-util/idToken');
+
 const router = express.Router();
 
 // ================ API router middleware: ================ //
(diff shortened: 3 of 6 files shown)


server/api/auth/createUserWithIdp.js
```diff
@@ -45,9 +45,14 @@ module.exports = (req, res) => {
...
 
+  console.log('Req.body', req.body);
   // Choose the idpClientId based on which authentication method is used.
   const idpClientId =
-    idpId === FACEBOOK_IDP_ID ? FACBOOK_APP_ID : idpId === GOOGLE_IDP_ID ? GOOGLE_CLIENT_ID : null;
+    idpId === FACEBOOK_IDP_ID
+      ? FACBOOK_APP_ID
+      : idpId === GOOGLE_IDP_ID
+      ? GOOGLE_CLIENT_ID
+      : 'client-id';
```
Gnito, line 55: If this 'client-id' string is meaningful variable for generic OpenId setup, maybe it should also be a constant with name that better communicates that it is.

server/api/auth/loginWithIdp.js
```diff
@@ -129,6 +129,6 @@ module.exports = (err, user, req, res, clientID, idpId) => {
         }
       );
 
-      res.redirect(`${rootUrl}${defaultConfirm}#`);
+      res.redirect(`${rootUrl}/confirm#`);
```
Gnito, line 132: Why has this been reverted to be a hard-coded endpoint?

server/apiRouter.js
```diff
@@ -80,4 +83,15 @@ router.get('/auth/google', authenticateGoogle);
...
 
+// This endpoint is called when user wants to initiate authenticaiton with Linkedin
+router.get('/auth/linkedin', authenticateLinkedin);
+
+// This is the route for callback URL the user is redirected after authenticating
+// with Linkedin. In this route a Passport.js custom callback is used for calling
+// loginWithIdp endpoint in Flex API to authenticate user to Flex
+router.get('/auth/linkedin/callback', authenticateLinkedinCallback);
+
+router.get('/.well-known/openid-configuration', openIdConfiguration);
```
Gnito, line 94: It would be good to have some comment or even link somewhere, where one can read more about these routes that this setup requires.

server/api-util/idToken.js
```diff
@@ -0,0 +1,82 @@
...
+const { default: SignJWT } = require('jose/jwt/sign');
+
+const radix = 10;
+const PORT = parseInt(process.env.REACT_APP_DEV_API_SERVER_PORT, radix);
+const rootUrl = process.env.REACT_APP_CANONICAL_ROOT_URL;
+const useDevApiServer = process.env.NODE_ENV === 'development' && !!PORT;
+
+const issuerUrl = useDevApiServer ? `http://localhost:${PORT}/api` : `${rootUrl}/api`;
+
+const clientId = process.env.CUSTOM_OIDC_CLIENT_ID;
```
Gnito, line 12: What is this environment variable?

Gnito: I think the biggest change I'd request is to remove these implementation-specific environment variables away from this file. 
- if CUSTOM_OIDC_CLIENT_ID is needed in implementing a new OpenId client, define it before calling functions from this file (i.e. use currying if necessary `exports.createIdToken = clientId =>  user => {`)

server/api-util/idToken.js
```diff
@@ -0,0 +1,82 @@
...
+
+// These keys are used for signing the ID token (JWT)
+// When you store them to environment variables you should replace
+// any line brakes with '\n'.
+// You should also make sure that the key size is big enough.
+const rsaSecretKey = process.env.RSA_SECRET_KEY;
+const rsaPublicKey = process.env.RSA_PUBLIC_KEY;
```
Gnito, line 19: Similarly, as with CUSTOM_OIDC_CLIENT_ID, I think that these implementation specific keys/environment variables are a bit too deeply placed in the util file. How about, defining these in apiRouter? Something like this:

```js
const rsaSecretKey = process.env.RSA_SECRET_KEY;
const rsaPublicKey = process.env.RSA_PUBLIC_KEY;
router.get('/.well-known/jwks.json', jwksUri(rsaSecretKey, rsaPublicKey));
```

That way customizers wouldn't need to pay too much attention to this utility file, but only configure new routes in apiRouter.

server/api/auth/linkedin.js
```diff
@@ -0,0 +1,110 @@
...
+        email,
+        firstName,
+        lastName,
+        idpToken,
+        from,
+        defaultReturn,
+        defaultConfirm,
+      };
+
+      console.log('userData:', userData);
```
Gnito, line 74: Does this contain something that should not be saved to server logs in production?

server/api/auth/linkedin.js
```diff
@@ -0,0 +1,110 @@
...
+  passport.authenticate('linkedin', {
+    state: paramsAsString,
+  })(req, res, next);
+};
+
+// Use custom callback for calling loginWithIdp enpoint
+// to log in the user to Flex with the data from Linkedin
+exports.authenticateLinkedinCallback = (req, res, next) => {
+  passport.authenticate('linkedin', function(err, user) {
+    loginWithIdp(err, user, req, res, 'client-id', 'linkedin');
```
Gnito, line 108: I'm quite confused about how this 'client-id' string works.
In loginWithIdp.js, there's **CLIENT_ID** constant and **clientId** param and they are not the same. There probably should be some kind of renaming made to these client id variables and values.

server/api/auth/createUserWithIdp.js
OtterleyW: This was just for testing. I thought I removed the WIP commit with hard-coded values from the PR but it seems I forgot to push that change. In real life, you should do a similar check for your custom implementation e.g. LinkedIn like you do e.g. for Google, and add also the required constants.

server/api/auth/loginWithIdp.js
OtterleyW: Same thing here, this was just for testing but I didn't remove the WIP commit.

server/api-util/idToken.js
OtterleyW: This is the client id you set to your identity provider in Console. In practice, users are free to set them whatever they wish as it doesn't need to be e.g. same as the LinkedIn client id (but it could be). The only requirement is that this clientId matches the one in Console.

server/api/auth/linkedin.js
OtterleyW: My initial idea was that we are not going to keep the LinkedIn implementation in FTW so I cut some corners when I did testing. So this is one of those parts. But if we want to keep this in the repo, then this of course needs to be fixed. 

Anyway, this 'client-id' is just a hard-coded test value. This 'client-id' the same thing you should have in process.env.CUSTOM_OIDC_CLIENT_ID and it's the value that should match the client id you have in Console. Most likely that environment variable is named to match the identity provider you are using (LinkedIn, Instagram, your own system etc.).

Gnito: Makes sense 👍 

How about the renaming of things in loginWithIdp.js file? (And sorry, this should have been addressed already in previous PRs.) In that file, there is CLIENT_ID referring to Flex client id and "clientId" referring to IDP client id - and then code like `clientId: CLIENT_ID,` - so, I'd imagine that this could be pretty confusing for someone who tries to understand what happens under the hood.

server/api-util/idToken.js
Gnito: Thanks for the explanation. 👍 

I'd suggest that we try to avoid "clientId" variable name and use something more clear instead (maybe idpClientID or oidcClientId or something?)

I'd also push all the environment variables that are implementation-specific away from this utility file (as I mentioned in the previous comment).

```
// Usage: createIdToken(process.env.CUSTOM_OIDC_CLIENT_ID)
exports.createIdToken = idpClientId => user => {
// etc.
}
```

server/apiRouter.js
```diff
@@ -80,4 +83,15 @@ router.get('/auth/google', authenticateGoogle);
...
+router.get('/.well-known/jwks.json', jwksUri);
```
lyyder, line 95: Should these be exposed only if, for example, the RSA variables are configured?

server/api-util/idToken.js
```diff
@@ -0,0 +1,82 @@
...
+    email: profile.email,
+    email_verified: profile.emailVerified,
+  })
+    .setProtectedHeader({ alg: 'RS256' })
+    .setIssuedAt()
+    .setIssuer(issuerUrl)
+    .setSubject(userId)
+    .setAudience(clientId)
+    .setExpirationTime('1h')
+    .sign(privateKey);
```
lyyder, line 58: Did we have the `kid` value in the token header? I might have mentioned that it's not compulsory but maybe we could add it by default to the ID token. Even though it's one more random value for the user to decide, enforcing the `kid` value in the token and in JWK could be useful as key caching in the Flex API relies heavily on it.

server/api-util/idToken.js
```diff
@@ -0,0 +1,82 @@
...
+    id_token_signing_alg_values_supported: ['RS256'],
+  });
+};
+
+// Serves the RSA public key as JWK
+// this document is expected to be found from
+// api/.well-known/jwks.json endpoint as stated in discovery document
+exports.jwksUri = (req, res) => {
+  fromKeyLike(crypto.createPublicKey(rsaPublicKey)).then(jwkPublicKey => {
+    res.json({ keys: [{ alg: 'RS256', use: 'sig', ...jwkPublicKey }] });
```
lyyder, line 80: Same here about he `kid` attribute.

lyyder (changes requested):
👍

server/api/auth/linkedin.js
OtterleyW: This is a really good point and I think renaming that parameter makes sense. I added that change to this PR now.

server/api-util/idToken.js
OtterleyW: Yes, we had but then I removed it to test if it works without it (and it did). I didn't want to leave it there with any hardcoded value. But I think it makes sense to have it in the example so it there again. Now the key id is also read from environment variables so it can be basically any string user decides. I guess this is ok?

server/api-util/idToken.js
```diff
@@ -79,11 +79,28 @@ exports.openIdConfiguration = (req, res) => {
...
+ * @param {Array} list containing keys to be served in json endpoint
+ *
+ * // Serves the RSA public key as JWK
 // this document is expected to be found from
 // api/.well-known/jwks.json endpoint as stated in discovery document
-exports.jwksUri = (rsaPublicKey, keyId) => (req, res) => {
-  fromKeyLike(crypto.createPublicKey(rsaPublicKey)).then(jwkPublicKey => {
-    res.json({ keys: [{ alg: 'RS256', kid: keyId, use: 'sig', ...jwkPublicKey }] });
+ */
+exports.jwksUri = (signingAlg, keys) => (req, res) => {
```
lyyder, line 90: Would it make sense to also include `alg` in the keys that are passed. As the `alg` attribute is key-specific, it would technically allow exposing keys with multiple algorithms. Quite an edge case feature right now but it would then be analogous to the JWK spec as there the algorithm is defined in the key. What do you think?

server/api-util/idToken.js
```diff
@@ -79,11 +79,28 @@ exports.openIdConfiguration = (req, res) => {
...
+  const jwkKeys = keys.map(key => {
+    return fromKeyLike(crypto.createPublicKey(key.rsaPublicKey));
+  });
+
+  Promise.all(jwkKeys).then(resolvedJwkKeys => {
+    const jwkKeys = resolvedJwkKeys.map((resolvedKey, i) => {
+      const originalKey = keys[i];
+      return { alg: signingAlg, kid: originalKey.keyId, use: 'sig', ...resolvedKey };
+    });
+    res.json({ keys: jwkKeys });
```
lyyder, line 104: Could the index access be avoided if this was was written somehow like this:

```js
  const jwkKeys = keys.map(key => {
    return fromKeyLike(crypto.createPublicKey(key.rsaPublicKey))
       // reads alg from key
      .then(res => {
        return {alg: key.alg, kid: key.keyId, ...res};
      });
  });

  Promise.all(jwkKeys).then(resolvedJwkKeys => {
    res.json({ keys: resolvedJwkKeys });
  });
```

lyyder (commented):
👍

server/api-util/idToken.js
OtterleyW: This makes sense!

server/api-util/idToken.js
OtterleyW: Thanks for the example! I tried to make it work like this yesterday but I might have missed something and the index access was the only way I managed to generate the list correctly. However, now it works like this and I agree that this is a better approach.

PR #1385, opened 2020-12-04
Add timeout for geolocation.getCurrentPosition call

Gnito:


src/util/maps.js
```diff
@@ -104,12 +104,18 @@ export const userLocation = () =>
       return;
     }
 
+    const options = {
```
OtterleyW, line 107: Where are these options coming from? Why do we need e.g. enableHighAccuracy to be true? I think having some kind of code comment to explain would be nice

OtterleyW (approved):

src/util/maps.js
Gnito: Good point. 👍 

They actually come from here: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition
I thought that "enableHighAccuracy" is a better default for templates since using false gave me a position that was 1km off.

PR #1386, opened 2020-12-04
Show whole world if no bounds or center

Gnito:


OtterleyW (approved):
Remember to update the changelog

PR #1389, opened 2020-12-08
Remove react-google-maps dependency

Gnito:
react-google-maps library has been unmaintained for 3 years.

Google Maps integration will be operated without any react wrapper library from now on. 
We just need a couple of classes from Google Maps API:
- **OverlayView** (This is needed to render HTML on top of the map.) 
- **Marker** or 
- "Circle" for fuzzy coordinates (I actually ended up using **Polygon** to improve roundness)

OverlayView is definitely the most complex setup. Luckily, part of the work to support it without react-google-maps was already done with some old bug fixes that were made on top of *react-google-maps* and generalization made to support *Mapbox*.

Other notes:
- Renaming: SearchMapWithGoogleMap => SearchMapWithGoogleMaps


src/components/SearchMap/SearchMap.js
```diff
@@ -145,11 +145,8 @@ export class SearchMapComponent extends Component {
     // When changing from default map provider to Google Maps, you should use the following
     // component instead of SearchMapWithMapbox:
     //
-    // <SearchMapWithGoogleMap
-    //   containerElement={
-    //     <div id="search-map-container" className={classes} onClick={this.onMapClicked} />
-    //   }
-    //   mapElement={<div className={mapRootClassName || css.mapRoot} />}
+    // <SearchMapWithGoogleMaps
```
kpuputti, line 148: It might be useful to briefly mention the component name change in the PR description.

src/util/googleMaps.js
```diff
@@ -154,3 +154,141 @@ export const locationBounds = (latlng, distance) => {
...
+ */
+export const getLayoutStyles = (mapCanvasProjection, offset, props) => {
+  if (props.bounds) {
+    const bounds = ensureOfType(props.bounds, google.maps.LatLngBounds, createLatLngBounds);
+    return getLayoutStylesByBounds(mapCanvasProjection, offset, bounds);
+  } else {
+    const position = ensureOfType(props.position, google.maps.LatLng, createLatLng);
+    return getLayoutStylesByPosition(mapCanvasProjection, offset, position);
+  }
+};
```
kpuputti, line 294: I think the helper functions in this file are very specific, and only internal helpers for this single export. Could they be internal functions here to avoid too much confusion?

You should probably also link directly to a commit hash version in GitHub, not just the master version of the file. That link might not contain anything in the future.

kpuputti (commented):
I will submit this partial review already now for you to read. I still have to review the largest file.

src/components/SearchMap/SearchMapWithGoogleMaps.js
```diff
@@ -3,21 +3,29 @@ import ReactDOM from 'react-dom';
...
 export const LABEL_HANDLE = 'SearchMapLabel';
 export const INFO_CARD_HANDLE = 'SearchMapInfoCard';
 const BOUNDS_FIXED_PRECISION = 8;
 
+const OVERLAY_VIEW = 'overlayView';
+const FLOAT_PANE = 'floatPane';
+const MAP_PANE = 'mapPane';
+const MARKER_LAYER = 'markerLayer';
+const OVERLAY_LAYER = 'overlayLayer';
+const OVERLAY_MOUSE_TARGET = 'overlayMouseTarget';
```
kpuputti, line 27: If these constants are values defined in the Google Maps API, you should clarify that very explicitly here. At least add a comment to the API reference, maybe even prefix the vars with `GMAPS_` etc?

src/components/SearchMap/SearchMapWithGoogleMaps.js
```diff
@@ -83,11 +91,28 @@ export const isMapsLibLoaded = () =>
...
+    overlayView.onAdd = this.onAdd.bind(this);
+    overlayView.draw = this.draw.bind(this);
+    overlayView.onRemove = this.onRemove.bind(this);
+    this.onPositionElement = this.onPositionElement.bind(this);
+
+    // You must call setMap() with a valid Map object to trigger the call to
+    // the onAdd() method and setMap(null) in order to trigger the onRemove() method.
+    overlayView.setMap(props.map);
+    this.state = {
+      [OVERLAY_VIEW]: overlayView,
```
kpuputti, line 112: Why does the key have to be dynamic here? If anyone wants to add new state to this component, they have to look carefully what the actual value of `OVERLAY_VIEW` is to know what key is used in the state object.

src/util/googleMaps.js
Gnito: @kpuputti, could you elaborate more? There are just 2 new exports: `getOffsetOverride` & `getLayoutStyles`. Are you suggesting that I move those internal functions inside of the scope of `getLayoutStyles` function (or something else)?

src/components/SearchMap/SearchMapWithGoogleMaps.js
```diff
@@ -422,29 +418,77 @@ class SearchMapWithGoogleMap extends Component {
...
+  }
+
+  initializeMap() {
+    const { offsetHeight, offsetWidth } = this.state.mapContainer;
+    const hasDimensions = offsetHeight > 0 && offsetWidth > 0;
+
+    if (hasDimensions) {
+      const { bounds, center, zoom } = this.props;
+      const maps = window.google.maps;
+      const controlPosition = window.google.maps.ControlPosition.LEFT_TOP;
```
kpuputti, line 449:
```suggestion
      const controlPosition = maps.ControlPosition.LEFT_TOP;
```

src/components/SearchMap/SearchMap.js
```diff
@@ -182,13 +181,13 @@ export class SearchMapComponent extends Component {
...
           createURLToListing={this.createURLToListing}
+          onClick={this.onMapClicked}
           onListingClicked={this.onListingClicked}
           onListingInfoCardClicked={this.onListingInfoCardClicked}
           onMapLoad={this.onMapLoadHandler}
-          onClick={this.onMapClicked}
           onMapMoveEnd={onMapMoveEnd}
-          zoom={zoom}
           reusableMapHiddenHandle={REUSABLE_MAP_HIDDEN_HANDLE}
+          zoom={zoom}
```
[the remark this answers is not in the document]
Gnito: Yeah, I was a bit hesitant to make this change, but my assumption is that most of the customizers have not touched these props - the need for changes is most likely happening inside those _SearchMapPriceLabel_, _SearchMapGroupLabel_ and _SearchMapInfoCard_ components. And this is a good moment to refactor these setups a bit since the props are changing anyway and they are going to be closer to each other.

src/components/SearchMap/SearchMapWithGoogleMaps.js
```diff
@@ -422,29 +418,77 @@ class SearchMapWithGoogleMap extends Component {
...
+      const zoomAndCenter =
+        !bounds && !center ? { zoom: 1, center: { lat: 0, lng: 0 } } : { zoom, center };
```
kpuputti, line 451: Well that's a one-liner! I think this would be more clear the other way. You check if `zoom` and `center` are defined and use them, or you provide a default object:

```suggestion
        bounds && center
          ? { zoom, center }
          : { zoom: 1, center: { lat: 0, lng: 0 } };
```

However, that is not equivalent to the code in the PR. Is the logic correct? Do you really want to use `{ zoom, center }` if either one is defined?

src/components/SearchMap/SearchMapWithGoogleMaps.js
```diff
@@ -466,22 +510,54 @@ class SearchMapWithGoogleMap extends Component {
...
     return (
-      <MapWithGoogleMap
-        onMapLoad={this.onMapLoad}
-        onIdle={this.onIdle}
-        bounds={bounds}
-        {...zoomAndCenter}
-        {...rest}
-      />
+      <div
+        id="map"
```
kpuputti, line 531: I think this is a bit too prone to conflict. If the id is really required, could it come as a prop from the page?

src/components/SearchMap/SearchMapWithGoogleMaps.js
```diff
@@ -490,8 +566,8 @@ SearchMapWithGoogleMap.defaultProps = {
   reusableMapHiddenHandle: null,
 };
 
-SearchMapWithGoogleMap.propTypes = {
-  containerElement: node.isRequired,
+SearchMapWithGoogleMaps.propTypes = {
+  // containerElement: node.isRequired,
```
kpuputti, line 570: Remove line?

src/util/googleMaps.js
kpuputti: I mean getting rid of functions like `createLatLng`, `createLatLngBounds`, and `ensureOfType` that are only used in this export might make sense. They are not that useful as generic helpers.

Gnito: I'll move those inside `getLayoutStyles`, but I don't want to completely remove them since these are rarely-touched helpers - and I think they should be easy to map with functions found on the other side of the reference link to react-google-maps repo.

And I changed those link to point to the latest tag instead of master.

src/components/SearchMap/SearchMapWithGoogleMaps.js
Gnito: We use just 2 of them, so I left only those and added better code comments to provide some context what these are.

```js
// Panes on Google Maps specify the stacking order for different layers on the map.
// https://developers.google.com/maps/documentation/javascript/customoverlays#intitialize
//
// Google Maps uses 5 different panes:
// 'mapPane', 'overlayLayer', 'markerLayer', 'overlayMouseTarget', 'floatPane'
// We only need the last 2:
// - 'overlayMouseTarget': for the SearchMapPriceLabelWithOverlay & SearchMapGroupLabelWithOverlay.
// - 'floatPane': to render InfoCardComponent on top of price and group labels.
```

src/components/SearchMap/SearchMapWithGoogleMaps.js
Gnito: Good point! It's some old code from react-google-maps keeping an old bugfix closer to library's codebase for potential upstream updates. Removed.

src/components/SearchMap/SearchMapWithGoogleMaps.js
Gnito: When at least one of them (bounds or center) is defined, we want to use them:
- If bounds are defined => map should be fitted to bounds.
- If only the center is defined => map should show that place (i.e. origin) in given or default zoom-level (e.g. street address: Erottajankatu 19 )

Only in this specific case where no bounds or center exists, we default to the center of the world/Mercator projection and show it with most-zoomed out viewport => to show the whole earth.

Prettier gave this formatting, but since this seems to be hard to understand one-liner, I'll format it:
```js
const zoomOutToShowEarth = { zoom: 1, center: { lat: 0, lng: 0 } };
const zoomAndCenter = !bounds && !center ? zoomOutToShowEarth : { zoom, center };
```

src/components/SearchMap/SearchMapWithGoogleMaps.js
Gnito: I'd keep the id because Google Maps is quite difficult to debug without some `document.getElementById()` calls from DevTools. (The element on those map panes are not selectable by right-clicking on top of the map.)

Anyway, I passed the id ("searchMap") from SearchMap component, but I didn't push it further than that. (And anyway, it's unlikely that there's more than one SearchMap instance since it takes 2/5 of the entire screen.

kpuputti: Yes, it is unlikely that a page in the app will have multiple search maps. However, we do have plans to use the styleguide more for lots of things like map variants etc., so it would be helpful to be able to define unique IDs there.

Gnito: Showing these components, that take the entire-screen into use, needs some thinking.  However, the change I made now should allow different ids to be passed for the SearchMap component.

PR #1391, opened 2020-12-11
Metadata was not passed through sanitizeUser function

Gnito:
User generated content can be sanitized in src/util/sanitize.js 
This PR fixes a bug related to that file: `sanitizeUser` function didn't pass metadata through.


OtterleyW (approved):
LGTM

PR #1392, opened 2020-12-11
Handle entity update with sparse attributes

Gnito:
I think we have forgotten to think about this merging setup when sparse attributes were added to templates. 
However, I'm not entirely sure if we should make this update or not. Is there any reasonable customization concept that relies on "reduction" of marketplace data in the store?
(This PR was motivated by a recent question in Slack.)

### DIFF
--- a/src/util/data.js
+++ b/src/util/data.js
@@ -27,7 +27,10 @@ export const combinedResourceObjects = (oldRes, newRes) => {
     throw new Error('Cannot merge resource objects with different ids or types');
   }
   const attributes = newRes.attributes || oldRes.attributes;
-  const attrs = attributes ? { attributes: { ...attributes } } : null;
+  const attributesOld = oldRes.attributes || {};
+  const attributesNew = newRes.attributes || {};
+  // Allow (potentially) sparse attributes to update only relevant fields
+  const attrs = attributes ? { attributes: { ...attributesOld, ...attributesNew } } : null;
   const relationships = combinedRelationships(oldRes.relationships, newRes.relationships);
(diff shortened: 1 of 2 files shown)


src/util/data.js
```diff
@@ -27,7 +27,10 @@ export const combinedResourceObjects = (oldRes, newRes) => {
     throw new Error('Cannot merge resource objects with different ids or types');
   }
   const attributes = newRes.attributes || oldRes.attributes;
-  const attrs = attributes ? { attributes: { ...attributes } } : null;
+  const attributesOld = oldRes.attributes || {};
+  const attributesNew = newRes.attributes || {};
+  // Allow (potentially) sparse attributes to update only relevant fields
+  const attrs = attributes ? { attributes: { ...attributesOld, ...attributesNew } } : null;
```
kpuputti, line 33: I think this is fine. We really should always use the latest attributes we have already fetched.

This will result in part of the data being stale, while part being updated. Before we showed only fully stale or fully updated data. But I think this is perfectly ok. If the sparse updates don't make sense, e.g. in cases where extended data is synced along with other attributes, then more sparse attributes can always be added where the update happens, right?

kpuputti (approved):

src/util/data.js
Gnito: @kpuputti Yes. Mainly, I was thinking about the reduction of data. E.g. if a listing has 50kb worth of publicData that somehow cumulates in the long-living app. But I guess that's an edge-case scenario that needs to be addressed case-by-case in customizations. (and full page-refreshes do happen every now and then)

PR #1397, opened 2021-01-05
Make sure verify API call succeeded before redirecting from EmailVerificationPage

OtterleyW:
This PRs adds a check that the verify email API endpoint has been called successfully before redirecting the user away from EmailVerificationPage. 

Earlier, if another user was logged in that user had already verified their email the redirect happened even if the actual API call failed with 403 Forbidden. This caused some confusion if the user didn't notice they were logged in with the wrong account. 

Now we check the `isVerified` value which should be true only if the API call was successful. Checking just `verificationError ` was not enough because initially the `verificationError` is always null and the redirect happens before that value gets updated. Now if the user is logged in with the wrong account they will see the error that email verification failed.

### DIFF
--- a/src/containers/EmailVerificationPage/EmailVerificationPage.js
+++ b/src/containers/EmailVerificationPage/EmailVerificationPage.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { bool, func, shape, string } from 'prop-types';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/containers/EmailVerificationPage/EmailVerificationPage.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { bool, func, shape, string } from 'prop-types';
```
Gnito, line 2: 👍

Gnito (approved):
LGTM

PR #1399, opened 2021-01-12
Move OIDC .well-known/* routes outside apiRouter

OtterleyW:
Move well-known/\* endpoints related to OIDC proxy setup from `apiRouter`to new `wellKnownRouter` so that they can be enabled outside the basic auth setup (e.g. ins staging environment). It also makes it simpler to set the identity provider URL, because we can drop the `/api` part of the path. 

We need to add the new `wellKnownRouter` to both `server/index.js` and `server/apiServer.js` because these routes are no longer handled in `apiRouter` but we want to be able to use them with both `yarn run dev` and `yarn run dev-server` commands. 

This PR also renames the environment variable `RSA_SECRET_KEY` to `RSA_PRIVATE_KEY` for consistency.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -33,6 +33,7 @@ const sitemap = require('express-sitemap');
 const passport = require('passport');
 const auth = require('./auth');
 const apiRouter = require('./apiRouter');
+const wellKnownRouter = require('./wellKnownRouter');
 const renderer = require('./renderer');
 const dataLoader = require('./dataLoader');
 const fs = require('fs');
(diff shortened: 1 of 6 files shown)


lyyder:
Also remember to update `jwks.json` location to be served from the root in the discovery document in `api-util/idToken.js`

server/index.js
```diff
@@ -131,6 +132,12 @@ app.use('/static', express.static(path.join(buildPath, 'static')));
...
 
+// These .well-known/* endpoints will be enabled if you are using FTW as OIDC proxy
+// https://www.sharetribe.com/docs/cookbook-social-logins-and-sso/setup-open-id-connect-proxy/
+// We need to handle these endpoints separately so that they are accessible by Flex
+// even if you have enabled basic authentication e.g. in staging environment.
+app.use('/.well-known', wellKnownRouter);
+
 // Use basic authentication when not in dev mode. This is
 // intentionally after the static middleware to skip basic auth for
 // static resources.
```
lyyder, line 143: This could now mention that this is after static middleware and /.well-known endpoints as those will bypass basic auth.

lyyder (approved):

PR #1402, opened 2021-01-21
Note about omitting trailing slash

Gnito:


scripts/config.js
```diff
@@ -228,7 +228,7 @@ const advancedSettings = settings => {
       name: 'REACT_APP_CANONICAL_ROOT_URL',
       message: `What is your canonical root URL?
 ${chalk.dim(
-  'Canonical root URL of the marketplace is needed for social media sharing and SEO optimization. When developing the template application locally URL is usually http://localhost:3000'
+  'Canonical root URL of the marketplace is needed for social media sharing, SEO optimization, and social logins. When developing the template application locally URL is usually http://localhost:3000 (Note: you should omit any trailing slash)'
```
OtterleyW, line 231: REACT_APP_CANONICAL_ROOT_URL is also used in login as functionality, but I'm not sure if it's really useful to list all possible use cases here.

OtterleyW (approved):
One comment, but LGTM!

scripts/config.js
Gnito: @OtterleyW yeah, I ended up using the same phrase that Hannu did for the Flex Docs article.

Personally, I think that we should at some point consider renaming the variable, because its usage has change dramatically over the years. Maybe something like "REACT_APP_HOST_DOMAIN"

PR #1409, opened 2021-02-04
google maps

gilfar77:


OtterleyW:
@gilfar77 this seems to be an accidental PR from a customization project so I'm closing it for now.

PR #1416, opened 2021-02-23
Add LoadableComponentErrorBoundary for handling ChunkLoadErrors

OtterleyW:
Adds ErrorBoundary for showing the error page if there's a ChunkLoadError

Screenshot 2021-03-03 at 11 04 21


src/components/LayoutSingleColumn/LayoutSingleColumn.js
```diff
@@ -11,7 +11,7 @@ import css from './LayoutSingleColumn.module.css';
 
 const prepareChildren = children => {
   const childrenCount = React.Children.count(children);
-  if (!(childrenCount === 3 || childrenCount === 4)) {
+  if (!(childrenCount === 2 || childrenCount === 3)) {
```
OtterleyW, line 14: I guess this has been a copy paster error from LayoutSideNavigation? Before the LoadableComponentErrorBoundaryPage we didn't have any page that would use LayoutSingleColumn without footer so that's why this has maybe not been noticed.

src/components/LoadableComponentErrorBoundary/LoadableComponentErrorBoundaryPage.js
```diff
@@ -0,0 +1,61 @@
...
+  Logo,
+} from '../../components';
+
+import css from './LoadableComponentErrorBoundary.module.css';
+
+export const LoadableComponentErrorBoundaryPage = () => {
+  const landingPagePath = pathByRouteName('LandingPage', routeConfiguration());
+  const handleOnClick = () => {
+    if (typeof window !== 'undefined') {
+      window.location = landingPagePath;
```
OtterleyW, line 19: We want the app to make full page refresh when navigating back LandingPage. I'm not sure if there is some other (better) way to trigger that or if this is ok.

src/components/LayoutSingleColumn/LayoutSingleColumn.js
Gnito: Change also that ”Menu” on next line :)

src/components/LoadableComponentErrorBoundary/LoadableComponentErrorBoundaryPage.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+  };
+
+  const landingPageLink = (
+    <InlineTextButton onClick={handleOnClick}>
+      <FormattedMessage id="LoadableComponentErrorBoundaryPage.landingPageLink" />
+    </InlineTextButton>
+  );
+
+  const topbar = (
```
Gnito, line 29: I think you can move this code inside LayoutWrapperTopbar - no need to keep it in different variable.

src/components/LoadableComponentErrorBoundary/LoadableComponentErrorBoundaryPage.js
Gnito: I think this is fine.

src/components/LoadableComponentErrorBoundary/LoadableComponentErrorBoundary.module.css
```diff
@@ -0,0 +1,97 @@
+@import '../../styles/customMediaQueries.css';
```
Gnito, line 1: I don't think that customMediaQueries.css import is needed. 
This is not a separate chunk (the code in components/index.js end up to main-chunk).

PR #1425, opened 2021-03-18
Currency's upper-case conversion should not expect that env-variable is set

Gnito:


OtterleyW (approved):
👍 Maybe you could update the PR title to something more describing and move the current title to description :)

PR #1429, opened 2021-04-09
Fix: LoadableComponentErrorBoundary should be used in prod, not in dev

Gnito:
LoadableComponentErrorBoundary helps in situations, where production build changes in the server and long-living client app tries to fetch code chunks that don't exist anymore.

Note: in development mode with Hot Module Reloading (HMR) in use, this can actually cause error loops with message:
_"React will try to recreate this component tree from scratch using the error boundary you provided"_


OtterleyW (approved):
👍

PR #1441, opened 2021-05-20
Fix searchMode in new URL after map is changed

anh-dinh-jh:
This PR is used to:
- Keep the existing applied searchMode in a newly created URL when the bounds of a map is changed on SearchPage


Gnito:
@anh-dinh-jh nice catch and thanks for the pull request! 
I used your commit, but added couple of extra commits in a separate PR.
(I used Prettier formatting and added a changelog entry.) 

https://github.com/sharetribe/ftw-daily/pull/1443
I'm closing this PR in favor of that one.

PR #1448, opened 2021-06-08
Update package.json

larrywho21:


Gnito:
I assume that this was accidental pull request. 
We primarily support UNIX-like environments since web hosting services mostly work like that. So, "export" is the default - but indeed, Windows would need "set". 

Anyway, I'm closing this PR for now.

PR #1452, opened 2021-07-07
Fix PriceFilter popup not close when click outside on Safari

chanthientukk9:
This PR is used to:
- Fix PriceFilter popup not close when click outside on Safari. It happens because onBlur of &lt;div&gt; tag not working on Safari.

### DIFF
--- a/src/components/PriceFilter/PriceFilterPopup.js
+++ b/src/components/PriceFilter/PriceFilterPopup.js
@@ -6,6 +6,7 @@ import { propTypes } from '../../util/types';
 import { formatCurrencyMajorUnit } from '../../util/currency';
 import config from '../../config';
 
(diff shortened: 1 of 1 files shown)


Gnito:
Hi, and sorry - we haven't followed PRs during summer time :( 

It does make sense to use OutsideClickHandler in here. It's a bit newer component than PriceFilter itself - so, we have just not remembered to change it at the  time. (Actually, PriceFilter should get some extra attention at some point - there just always seem to be something else in the pipeline.)

Anyway, I'll get back to this hopefully at some point this week, but at least one change should be made: The comment inside `handleBlur` should be removed since it's not relevant after the change.)

Gnito:
Closing this in favour of https://github.com/sharetribe/ftw-daily/pull/1455

PR #1473, opened 2021-09-21
Increase input font size to avoid zoom on iOS

kpuputti:
If the input font size is smaller than 16px, iOS zooms in the page when the input gets focus. This PR changes the font size from 14px to 16px on mobile to avoid extra zooming.

## Screenshot with the current 14px font size

input-14px

When focusing on the input:

input-14px-focus

## Screenshot with the proposed 16px font size

input-16px


Gnito (approved):
LGTM. Changelog missing. 
(Perhaps the other PR should be merged first.)

PR #1476, opened 2021-10-04
Fix modal close button text/icon alignment

kpuputti:
This PR fixes the vertical alignment of the modal close button text and icon.

## Before

close-before

## After

close-after


Gnito (approved):
LGTM

PR #1483, opened 2021-11-29
Update dependencies

kpuputti:
Update several dependencies. See changes to `package.json` for full info.


Gnito:
Couple of comments: 
- **dotenv** libs should be in sync with Create React App's libs
- **Prettier**: I'm not sure if it's worth it since this will a) make it harder for customizers to take upstream update and b) we need to make upstream updates to FTW-hourly & FTW-daily

I didn't mention yesterday, but **loadable components** are also tied with our CRA fork - so, if sharetribe-scripts is not updated at the same time, then some SSR testing is needed to verify that client app versions of those libs are working with the version of loadable stuff we have on sharetribe-scripts.
https://github.com/sharetribe/create-react-app/blob/master/packages/react-scripts/package.json#L33-L35

kpuputti:
@Gnito Good points about Prettier. That just introduces conflicts with no real benefits. And customizers can easily update the lib and format if they really want that (although that might introduce conflicts in any case).

I did check to match `dotenv` with the versions in `sharetribe-scripts`, but good to know that Loadable should be matches as well. I'll check that.

kpuputti:
Now the Prettier and Loadable updates have been removed.

Gnito (approved):
LGTM. 
Note: it might make sense to review the other PR first before merging this to master. Mainly, because, that feature branch is deployed to staging atm.

https://github.com/sharetribe/ftw-daily/pull/1486

PR #1486, opened 2021-12-01
Fix checkout process

Gnito:
Issues:
  - Submit button was enabled prematurely for onetime payments. 
    It didn't check if stripeElements aka card number field was filled.
  - Changing from onetime payment to default card was not working correctly in case of error (e.g. network error).
  - Calling _stripe.confirmCardPayment_ when status is _requires_capture_ is unnecessary.
    >  400 "You cannot confirm this PaymentIntent because it has a status of requires_capture. Only a PaymentIntent with one of the following statuses may be confirmed: requires_confirmation, requires_action."
    

_Note_: these are cherry-picked commits from FTW-product:
  - [PR 111](https://github.com/sharetribe/ftw-product/pull/111)
  - [PR 120](https://github.com/sharetribe/ftw-product/pull/120)
  - [PR 121](https://github.com/sharetribe/ftw-product/pull/121)

    (Changelog updates were not included)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2021-XX-XX
 
+- [fix] Fix bugs in checkout process:
+
+  - Submit button was enabled prematurely for onetime payments
+  - Toggling between default card and onetime payment flows was not working correctly in case of
+    error (e.g. network error).
+  - Calling Stripe.confirmCardPayment when status is requires_capture is unnecessary.
+
+  [#1486](https://github.com/sharetribe/ftw-daily/pull/1486)
+
 - [change] Update many dependencies. See full list in the package.json changes in the PR.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,15 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2021-XX-XX
 
+- [fix] Checkout process had bugs:
```
kpuputti, line 17: Tiny thing, but this just tells that there were bugs, not that those bugs were fixed. :D

Maybe something like:

```suggestion
- [fix] Fix bugs in checkout process:
```

kpuputti (approved):
I don't fully understand the Stripe/PI flow, but the code looks clean in general.

CHANGELOG.md
Gnito: It felt strange to write `[fix] Fix`

PR #1491, opened 2021-12-31
Update README.md

shareoc:
Windows users commonly report the following error message when running `yarn run dev` or `dev-server`:
`RangeError [ERR_SOCKET_BAD_PORT]: options.port should be >= 0 and < 65536`

Setting multiple environment variables in Windows requires using the set command before each individual variable. Updated the 'for windows users' section in documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2022-XX-XX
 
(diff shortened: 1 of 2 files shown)


SariSaar (approved):

Gnito:
@shareoc 👍  
You should add a changelog entry and then this is ready to be merged. 
Newest entry first: `- [fix] ...`
And then Prettier: "yarn run format-docs"

shareoc:
@Gnito Added entry to changelog!

PR #1504, opened 2022-04-01
Testing new default location

tedserdel:
Updated the first default location to show the neighborhood of Capitol Hill in Seattle. I've used the template to get the LatLng of the desired neighborhood. I've zoomed out once for the map to show more context.


tedserdel (commented):
This looks good

PR #1557, opened 2022-10-23
Pages fixes

jannekoivistoinen:
- Finalize Dark Theme Support
- Hide container DIV's without content with :empty and {display: none}
- Fine-tune vertical spacing between multiple elements with adjacent sibling combinators for now, as support for :has is still lacking in Firefox by default.


src/containers/LandingPage/FallbackPage.js
```diff
@@ -7,69 +7,16 @@ export const fallbackSections = {
...
-            type: 'internalButtonLink',
-            href: '/s',
-            label: 'Browse marketplace',
-          },
-        },
-      ],
+      sectionId: 'maintenance-mode',
+      title: {
+        type: 'heading1',
+        content: 'HTTP 500 Internal Server Error',
```
Gnito, line 13: Ei tää ole Internal server error.

Kirjoita mieluummin jotain tyyliin:
"Maintenance mode" / "The marketplace is not fully operational at the moment." 
Ja ehkä vielä: "Try refreshing the page and if that doesn't solve the issue, contact the marketplace admins."

src/containers/PageBuilder/Primitives/Heading/Heading.module.css
```diff
@@ -6,24 +6,93 @@
...
+  & + ul,
+  & + ol,
+  & + code,
+  & + div,
+  & + h1,
+  & + h2,
+  & + h3,
+  & + h4,
+  & + h5,
+  & + h6 {
```
Gnito, line 31: In what situation, the margin-top is not used?

Could we turn this around? `margin-top: 24px;` is the default and it's not there in certain situations.

src/containers/PageBuilder/Primitives/Heading/Heading.module.css
```diff
@@ -6,24 +6,93 @@
...
```
Gnito, line 51: In what situation, the margin-top is not used?

Could we turn this around? `margin-top: 16px;` is the default and it's not there in certain situations.

src/containers/PageBuilder/Primitives/Heading/Heading.module.css
```diff
@@ -6,24 +6,93 @@
...
```
Gnito, line 71: In what situation, the margin-top is not used?

Could we turn this around? `margin-top: 8px;` is the default and it's not there in certain situations.

src/containers/PageBuilder/Primitives/Heading/Heading.module.css
Gnito: I found this:
https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child

So, maybe this could be written as:

```
.h1 {
  font-size: 40px;
  margin: 24px;

  &:only-child {
    margin: 0px;
  }
}
```

And the same approach to other situations too.

src/containers/PageBuilder/SectionBuilder/SectionBuilder.example.js
```diff
@@ -63,7 +63,7 @@ export const SectionArticle = {
               content: `Donec ullamcorper nulla non metus auctor fringilla. Cras justo odio, dapibus ac facilisis in, egestas eget quam. Etiam porta sem malesuada magna mollis euismod. Maecenas sed diam eget risus varius blandit sit amet non magna.
 Maecenas faucibus mollis interdum. Sed posuere consectetur est at lobortis. Etiam porta sem malesuada magna mollis euismod. Etiam porta sem malesuada magna mollis euismod. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Maecenas faucibus mollis interdum.
               
-Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum id ligula porta felis euismod semper. Nullam id dolor id nibh ultricies vehicula ut id elit. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
+[Lorem ipsum dolor sit amet](#), consectetur adipiscing elit. Vestibulum id ligula porta felis euismod semper. Nullam id dolor id nibh ultricies vehicula ut id elit. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
```
Gnito, line 66: Let's not add links to nowhere. 
If link is needed, then:
`[Lorem ipsum dolor sit amet](#cms-features-block-1)`

src/containers/PageBuilder/SectionBuilder/SectionBuilder.module.css
```diff
@@ -6,12 +6,29 @@
...
+  padding: 0 32px;
   position: relative;
 
   @media (--viewportMedium) {
     justify-content: center;
   }
+
+  & + div {
+    padding: 64px 32px 0;
+  }
```
Gnito, line 18: Define this before nested media query (and use :only-child)

Gnito: And don't use 3 argument version. Prefer: 2 or 4 for the sake of readability (most ppl/customizers are not familiar with [top] [left&right] [bottom] syntax).

In this case: padding: `64px 32px 0 32px;`

src/containers/PageBuilder/SectionBuilder/SectionBuilder.module.css
```diff
@@ -6,12 +6,29 @@
...
+
+.sectionDetails {
+  & h1 + *,
+  & h2 + *,
+  & h3 + *,
+  & h4 + *,
+  & h5 + *,
+  & h6 + * {
+    margin-top: 12px;
+  }
```
Gnito, line 31: Don't declare .sectionDetails multiple times.
=> move these lines inside previous ".sectionDetails"
=> don't use wildcard/asterisk if you can avoid it.
=> keep the custom media query as the last thing inside class rules

```
.sectionDetails {
  max-width: var(--contentMaxWidth);
  display: grid;
  justify-content: start;
  margin: 0 auto;
  padding: 64px 32px 0 32px;
  position: relative;

  &:only-child {
    padding: 0 32px;
  }
  
  /* Reduce the white-space after the sectionTitle a bit more than in the Blocks */
  & h1 + *,
  & h2 + *,
  & h3 + *,
  & h4 + *,
  & h5 + *,
  & h6 + * {
    margin-top: 12px;
  }
  
  @media (--viewportMedium) {
    justify-content: center;
  }
}
```

src/containers/PageBuilder/Primitives/P/P.module.css
```diff
@@ -2,10 +2,10 @@
...
   line-height: 1.66;
   margin-top: 0;
-  margin-bottom: 24px;
+  margin-bottom: 0;
   letter-spacing: 0;
-}
 
-h2 + p {
-  margin-top: 16px;
+  & + * {
```
Gnito, line 8: Try to use :only-child - aka swap it to this mode:
```
  margin-top: 24px;

  &:only-child {
    margin-top: 0;
  }
```

src/containers/PageBuilder/Primitives/List/List.module.css
```diff
@@ -14,9 +14,16 @@
...
+
+  & + * {
+    margin-top: 24px;
+  }
 }
 
 .ol {
+  & + * {
+    margin-top: 24px;
+  }
```
Gnito, line 26: What does this mean?
A list (ul & ol) should not be used unless there's list-item (li) inside.
Did you find a bug?

src/containers/PageBuilder/Primitives/List/List.module.css
```diff
@@ -14,9 +14,16 @@
       list-style-type: square;
     }
   }
+
+  & + * {
+    margin-top: 24px;
+  }
```
Gnito, line 20: What does this mean?
A list (ul & ol) should not be used unless there's a list-item (li) inside.
Did you find a bug?

src/containers/PageBuilder/Primitives/Ingress/Ingress.module.css
```diff
@@ -2,6 +2,10 @@
   font-size: 18px;
   line-height: 1.66;
   margin-top: 0;
-  margin-bottom: 24px;
+  margin-bottom: 0;
   letter-spacing: 0;
+
+  & + * {
```
Gnito, line 8: Use :only-child instead

src/containers/PageBuilder/Primitives/Link/Link.module.css
```diff
@@ -1,3 +1,8 @@
 .link {
+  display: inline-block;
   color: var(--marketplaceColor);
+
+  & + * {
```
Gnito, line 5: Use :only-child instead

src/containers/PageBuilder/SectionBuilder/SectionBuilder.module.css
```diff
@@ -61,11 +78,19 @@
...
+    background-color: cyan;
+    color: unset;
+  }
+}
+
 .darkTheme p,
 .darkTheme li,
 .darkTheme blockquote {
-  color: var(--matterColorLight);
+  color: rgba(255, 255, 255, 0.85);
```
Gnito, line 93: Is the opacity necessary? I.e. do you want the potential background image to be visible through text color?
But if you just want to use grey color, use grey color.

src/containers/PageBuilder/SectionBuilder/SectionBuilder.example.js
jannekoivistoinen: Ah yeah this was just for testing purposes as we didn't have any markdown links! I'll change it to an anchor link

src/containers/PageBuilder/Primitives/Heading/Heading.module.css
jannekoivistoinen: A primitive shouldn't have any whitespace around it on it's own, because that messes up the vertical spacing. Few examples,

- Sections always have margin-y of 64px.
- H1 has a margin-bottom of 24px. Section only has a H1, but the margin-bottom of the H1 gets added to the section's margin-bottom, which causes too much whitespace below the H1, 88px which doesn't match with the margin-top. Same thing with primitive default margin-tops, too much whitespace.
- Headings followed by another headings might require more margin-bottom than for example paragraphs or buttons
- So the common solution is to use sibling selectors and instead of building the margins down for each primitive, we build margins up after the first primitive per the whitespace relationship between primitives.

The infamous :has selector which I've been trying to get your opinion on solves all the pain for "context-aware" margins between primitives. Not sure if :only-child will do the trick, but I'll try!

src/containers/LandingPage/FallbackPage.js
Gnito: You could fine-tune this more. E.g. use an image or something in that direction.
For the Sharetribe-hosted end-users, it will communicate that the marketplace is not fully functional at the moment.
(This should be a very rare event, but it's still possible that assets are not working properly.)

src/containers/PageBuilder/Primitives/List/List.module.css
jannekoivistoinen: @Gnito it means that if there's an element after the List component, the element gets a margin-top and that determines the white-space between the List component and the element. If there's no element after the List component, there won't be a white-space as it's not necessary.

PR #1561, opened 2022-11-16
Fix carousel

Gnito:
Fix carousel width calculation and add responsive behaviour.

### DIFF
--- a/src/containers/PageBuilder/SectionBuilder/SectionCarousel/SectionCarousel.js
+++ b/src/containers/PageBuilder/SectionBuilder/SectionCarousel/SectionCarousel.js
@@ -1,5 +1,6 @@
 import React, { useEffect } from 'react';
 import { arrayOf, func, node, number, object, shape, string } from 'prop-types';
+import classNames from 'classnames';
 
 import Field, { hasDataInFields } from '../../Field';
 import BlockBuilder from '../../BlockBuilder';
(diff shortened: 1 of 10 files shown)


src/containers/PageBuilder/SectionBuilder/SectionCarousel/SectionCarousel.module.css
```diff
@@ -4,10 +4,16 @@
...
   align-items: center;
   scroll-behavior: smooth;
+  padding: 64px 0 0 0;
 
-  @media (--viewportMedium) {
+  &:hover .carouselArrows {
+    opacity: 1;
+  }
+
+  /* smartphones, touchscreens: we don't need to show arrows */
```
jannekoivistoinen, line 13: @Gnito how will this work with devices that have a keyboard, mouse/trackpad and a touch screen? Some laptops have touch screens. Latest iPads are also going to this direction with their Magic Keyboards etc.

src/containers/PageBuilder/SectionBuilder/SectionCarousel/SectionCarousel.js
```diff
@@ -109,8 +113,12 @@ const SectionCarousel = props => {
         </header>
       ) : null}
       {hasBlocks ? (
-        <div className={css.carouselContainer}>
-          <div className={css.carouselArrows}>
+        <div className={css.carouselContainer} id={sliderContainerId}>
```
jannekoivistoinen, line 116: Nice one! 🍬

src/containers/PageBuilder/SectionBuilder/SectionCarousel/SectionCarousel.module.css
Gnito: @jannekoivistoinen Well, it's true that this follows the primary input mechanism. So, this needs a bit more testing, but to me, it feels like hover effects on top of the content in this particular scenario are pretty annoying on touch devices. Maybe more annoying than the benefit of showing the arrows if there are secondary pointing inputs that support hover. 

Note: 
since it's obvious from the design that there is more content, a user is likely to try dragging the scrollable pane anyway - which should work. Another way that this should work is by clicking the scrollable pane and then using the arrow keys.

But good question! Maybe I should have used "any-hover: none". (I wasn't even aware of it.)

Issue #1601, opened 2023-11-28
The function 'routeConfiguration' does not work.


deepakrh185:
**When attempting to access the current location using the build browser navigator function, an error is thrown stating that navigator is undefined. To resolve this issue, the following code was added: `if (typeof navigator !== 'undefined')`. However, after implementing this fix, the application now throws an error stating that routeConfiguration is not a function**

![image](https://github.com/sharetribe/ftw-daily/assets/71048565/b198aa35-1e8b-4167-bad6-548a6985f05a)

**Steps to reproduce the behaviour:**
1. Create a new file called useCurrentLocation.js with this code
```
import { types as sdkTypes } from '../util/sdkLoader';
const { LatLng } = sdkTypes;

export const useCurrentLocation = async () => {
  if (typeof navigator !== 'undefined') {
    if (!('geolocation' in navigator)) {
      throw new Error('Geolocation not available in the browser');
    }

    const options = {
      enableHighAccuracy: true,
      timeout: 5000,
      maximumAge: 0,
    };

    try {
      const position = await new Promise((resolve, reject) => {
        navigator?.geolocation?.getCurrentPosition(resolve, reject, options);
      });

      return new LatLng(position.coords.latitude, position.coords.longitude);
    } catch (error) {
      console.log('error->>>Location', error);
      return error.message;
    }
  } else {
    return;
  }
};
````
2.Call the function in Searchpage.duck.js to access the current Location
```
const { useCurrentLocation } = await import('../../hooks/useCurrentLocation');
 const getCurrentLocation = useCurrentLocation();
 async function fetchResults() {
  const latlng = await getCurrentLocation;
  console.log(latlng);
  const results = {
    address: '',
    origin: latlng,
    bounds: locationBounds(
      latlng == 'User denied Geolocation' ? null : latlng,
      config.maps.search.currentLocationBoundsDistance
    ),
  };
  return results;
}
```

**Expected behavior** This should run without error when `npm run dev-server`

**Screenshots** 
![image](https://github.com/sharetribe/ftw-daily/assets/71048565/a60d59f3-e7d2-4a78-821f-3e0c54a9591e)

deepakrh185:
The solution to the problem can be found in this pull request: https://github.com/sharetribe/ftw-daily/pull/1602. @

Gnito:
We are archiving this repository, since we don't really have resources to keep these FTW-* templates up to date.

[...]

PR #1602, opened 2023-11-28
Fixed the function 'routeConfiguration' does not work

deepakrh185:
Fixes for #1601 @kpuputti   please review this PR

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -14,6 +14,7 @@
  */
 
 // This enables nice stacktraces from the minified production bundle
+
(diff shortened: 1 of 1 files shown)


Gnito:
We are archiving this repository, since we don't really have resources to keep these FTW-* templates up to date.

We'll recommend that new marketplaces should be created on top of web-template. And that's the best place to find implementation details if new features are interesting enough to be implemented on top of these older templates that are dedicated to single transaction process.

